Issue #1214, opened 2018-03-05
Support the formatID : http://www.isotc211.org/2005/gmd-pangaea


mbjones:
---

Author Name: **Jing Tao** (Jing Tao)
Original Redmine Issue: 7231, https://projects.ecoinformatics.org/ecoinfo/issues/7231
Original Date: 2017-12-07
Original Assignee: Jing Tao

---

The schema of the format id has been added and the db script to modify the xml_catalog table has been added also.
In metacat-index, the subporcessor for this format id has been added.

The change was committed to both trunk and 2.8 branch.

PR #1233, opened 2018-04-05
Various fixes for Pangaea ISO rendering

amoeba:
See the commits for more detailed information but the gist of this is that the Pangaea format ID wasn't hooked up in the view service when our initial support for Pangaea's format ID was added in #1214. So I linked it in. Then I found a rendering issue that affected Pangaea so I fixed that. Then I made a few more minor tweaks.

I've only done a little bit of testing but everything looks fine.


amoeba:
Since this never got merged, I merged this branch into https://github.com/NCEAS/metacat/tree/feature-improve-iso-view so I'll close this and open a new PR for that branch.

PR #1240, opened 2018-05-01
Improve ISO view for generic ISO, NCEI, and Pangaea content

amoeba:
This PR contains various bug fixes some major improvements to the ISO View Service XLTs. These changes will impact any ISO19139 content and how it renders out of the View Service.

Summary:

- Moved the Alternate Data Access Table from MetacatUI to the View Service https://github.com/NCEAS/metacatui/issues/583
- Added support for how PANGAEA does Parameters (attributes) by inserting a simple table of the parameters when the metadata has `contentInfo` elements in it
- Added support for how PANGAEA does links to their data downloads (`distributionInfo`) by putting a table of links at the top of the view
- Fixed various minor issues in the XSLTs which caused headers to get shown when there was no content in the metadata for that section

@taojing2002 would you mind taking a look-through? I've hot-patched `cn-stage-ucsb-1` with these changes and done some moderate testing but always appreciate feedbacl. See https://search-stage.test.dataone.org/#view/doi:10.1594/PANGAEA.832895_20170724_1121 for an example.

### DIFF
--- a/lib/style/skins/metacatui/iso19115/iso-md.xsl
+++ b/lib/style/skins/metacatui/iso19115/iso-md.xsl
@@ -188,4 +188,87 @@
         <xsl:value-of select="./*/gml:identifier/text()" />
         <xsl:value-of select="./*/gml:name/text()" />
     </xsl:template>
+    <!-- gmd:MD_Distribution 
+        - gmd:distributionFormat, 0-inf ( just grab gmd:MD_Format/gmd:name for each)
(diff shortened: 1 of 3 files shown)


amoeba:
Hey @taojing2002 this is code we would like in the next release but we'd also like this to get released with @laurenwalker 's MetacatUI 1.14.15 patch release. We moved the Alternate Data Access Table from MetacatUI and put it in Metacat's View Service so if these two products aren't released/patched at the same time the user will see two tables.

Issue #1242, opened 2018-05-01
Make ant build auto-detect maven location


amoeba:
`maven.home` in build.properties is currently hard-coded and it requires developers to change it to get the build running.

https://github.com/NCEAS/metacat/blob/1fbd3b87dc52cd77ee80ab376eea45db37a7cf9d/build.properties#L83

It seems like this can be refactored to automatically provide the build with the location of maven based up on the $PATH.

amoeba:
My first idea;

```bash
$ export MAVEN_HOME="$(mvn --version | grep "Maven home" | cut -d" " -f3)"
$ echo $MAVEN_HOME
/usr/local/Cellar/maven/3.5.3/libexec
```

Gotta be something better than that.

mbjones:
I think the algorithm should be:

1) use $MAVEN_HOME if it is set in the environment
2) if mvn is on the path, run it to find MAVEN_HOME as @amoeba proposed
3) if mvn is not in the path, see if it is in common locations (/usr/local/bin, /opt/local/bin, /opt/bin, etc), and if it is found, then extract MAVEN_HOME as in (2)
4) if still not found, fail build with message saying to do 1 ,2, or 3

amoeba:
I figured this out and it seems to work well. I skipped your Step 3 since I didn't think it was nearly as important as Steps 1 or 2 and thought Steps 1 and 2 were good enough.

Here it is in action:

```bash
$ sh find-maven.sh
A proper MAVEN_HOME could not be found. Either:

  - Set MAVEN_HOME i.e., export MAVEN_HOME="{SOMETHING}"
  - Make your mvn executable available in your PATH

And then run this script (or ant if that's the command you just ran
when you got this error.
$ export MAVEN_HOME="/usr/local/maven/some_home"
$ sh find-maven.sh
/usr/local/maven/some_home
$ export PATH="/usr/local/bin:$PATH"
$ unset MAVEN_HOME
$ sh find-maven.sh
/usr/local/Cellar/maven/3.5.3/libexec
```

And I've tested this out locally by producing a successful ant build. I sent this as a PR in case anyone wants to have a look first: https://github.com/NCEAS/metacat/pull/1245

PR #1245, opened 2018-05-08
Write a script to auto detect maven.home in the ant build

amoeba:
This was annoying to set every time I ran the ant build and it seemed like something that could be automated. I figured out how. Hopefully addresses #1242 satisfactorally.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -79,6 +79,3 @@ make.perl.code=true
 # The metacat current dir's abolute value. If you set install.ecogrid=false, this variable
 #  do NOT need to be configured
 #metacat.dir=/home/daigle/workspace/metacat
-
-#maven.home=/usr/local/Cellar/maven/3.5.0/libexec/
-maven.home=/usr/share/maven
--- a/build.xml
+++ b/build.xml
@@ -161,6 +161,12 @@
 			<redirector outputproperty="found.maven3.home" />
 		</exec>
 
+		<exec executable="sh" resultproperty="exec.mvn2.status" failifexecutionfails="false">
+			<arg value="-c"/>
+			<arg value="dpkg-query -L maven2 | grep '/boot$'" />
+			<redirector outputproperty="found.maven2.home" />
+		</exec>
+
 		<!-- Use value set in $MAVEN_HOME if maven.home isn't already set -->
 		<condition property="isset.env.MAVEN_HOME">
 			<isset property="env.MAVEN_HOME"/>
@@ -187,6 +193,20 @@
 
 		<!-- Fall back to the found value via the above <exex> mvn -->
 		<condition property="maven.home" value="${found.maven.home}">
+    		<and>
+				<not>
+					<isset property="maven.home" />
+				</not>
+				<not>
+					<!-- Helps this condition fall through to find maven 2, below -->
+					<equals arg1="${found.maven.home}" arg2="" />
+				</not>
+				<equals arg1="${which.mvn}" arg2="0" />
+			</and>
+		</condition>
+
+		<!-- Fall back to the found value via the above <exex> mvn when version 2 is installed -->
+		<condition property="maven.home" value="${found.maven2.home}">
     		<and>
 				<not>
 					<isset property="maven.home" />
--- a/build.xml
+++ b/build.xml
@@ -163,7 +163,7 @@
 
 		<exec executable="sh" resultproperty="exec.mvn2.status" failifexecutionfails="false">
 			<arg value="-c"/>
-			<arg value="dpkg-query -L maven2 | grep '/boot$'" />
+			<arg value="dpkg-query -L maven2 | grep '/boot$' | cut -d'/' -f1-4" />
 			<redirector outputproperty="found.maven2.home" />
 		</exec>
 
--- a/build.xml
+++ b/build.xml
@@ -47,6 +47,11 @@
              uri="antlib:org.apache.maven.artifact.ant"
              classpathref="maven-ant-tasks.classpath" />
 	
+	<!--  Don't run setMavenHome target if maven.home is already set -->
+	<condition property="maven.home.isset">
+		<isset property="maven.home" />
+	</condition>
+
 	<!-- defining the axis tasks -->
 	<path id="axis.ant.classpath">
 		<fileset dir="lib/lsid_lib">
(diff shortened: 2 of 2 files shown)


mbjones:
Awesome. Glad it was improved. Certainly a huge step up from the build property, which I agree was so annoying.

Does it work on windows?  Seems like the same functional tests could have been implemented in Ant rather than bash -- first use Ant to test if the env var exists, and if not, then use ant's `exec` task to start the mvn executable and capture the output to set the MAVEN_HOME env var.  Its probably good enough to commit this as is, but it would be nice to keep the Windows build working. Let's discuss.

amoeba:
> Does it work on windows? Seems like the same functional tests could have been implemented in Ant rather than bash

I'm pretty sure it wouldn't work (outside of WSL) since I hardcode `/bin/sh`. Doing this in Ant is beyond me at the moment so doing this via a shell script was a lot faster for me. I'd be happy to see this done entirely in Ant so we can support Windows.

I don't think I have the time to invest right now to learn enough Ant to make this work. Maybe somebody else could bang this out in Ant real quick?

taojing2002:
In the find-maven.sh script file, is it possible to handle the command `mvn3`? In my mac, I only have the `mvn3` command.

amoeba:
I don't see why not @taojing2002. Just curious, what install method gave you a `mvn3` executable?

amoeba:
@taojing2002 thanks 👍 . What did you think about @mbjones 's idea to in-line this into the ant build rather than putting this in as separate file? Any chance you're experienced writing Ant tasks?

amoeba:
I decided to convert my previous shell script into a proper ant ask. I force-pushed the branch and re-based it off of `dev-2.10` so any code you saw before is gone.I think I like ant more than I expected.

@taojing2002 can you give this a whirl some time? It should just work for you out of the box so long as the shell you call `ant` from can see a `mvn3` executable. You should then be able to override the value by adding `maven.home` to `build.properties` or setting a `$MAVEN_HOME` env var.

The new behavior is:

- Trust `maven.home	`
- Fall back to `$MAVEN_HOME`
- Fall back to asking `mvn3` and then `mvn` to print its home to stdout and capture it
- Fail the task if `maven.home` isn't set to something after all of this

taojing2002:
This is great. However, when I tried this branch on mn-demo-6.test.dataone.org,  which we always use as a test server, it didn't work. I think other demo machines don't work either. Is it possible to make this case work out of box?

amoeba:
Thanks for checking it out. 

> Is it possible to make this case work out of box?

Most definitely. 

Re:

> it didn't work

What behavior do you see? What kind of system is that?

taojing2002:
The mn-demo-6 was just upgraded to Ubuntu 18.04 and the maven was removed. After I reinstall maven, it works. My bad.

However, the branch doesn't work with Ubuntu 14.04 on mn-sandbox-ucsb-1. The error message is:
`BUILD FAILED
/home/dev/tao/project/metacat/build.xml:1712: /home/dev/tao/project/metacat/boot does not exist.`
The maven version is:
`tao@mn-sandbox-ucsb-1:~/project/metacat$ mvn -version`
`Apache Maven 2.2.1 (rdebian-14)
`

amoeba:
> The mn-demo-6 was just upgraded to Ubuntu 18.04 and the maven was removed. After I reinstall maven, it works. My bad.

Did the setMavenHome task fail with a helpful error message or fail in a weird way?

> However, the branch doesn't work with Ubuntu 14.04 on mn-sandbox-ucsb-1. The error message is:
BUILD FAILED /home/dev/tao/project/metacat/build.xml:1712: /home/dev/tao/project/metacat/boot does not exist.

Weird. Any chance I can get access to that box? You don't have a `$MAVEN_HOME` var set do you?

taojing2002:
I don't have a `MAVEN_HOME` setting. Would you please ask Nick to create an account on that machine for you?

amoeba:
Sure 👍

amoeba:
This is caused behavior that Maven 2 (end of lifed in 2009). So I expect this ant target will work for maven 3 and newer. I'll see if I can find a nice way of handling Maven 2.

amoeba:
@taojing2002 I pushed a fix and tested on mn-sandbox-1 and it seems to work. Thanks for catching that edge case.

taojing2002:
Really? I just pulled the update and ran again, but still got an error:
`BUILD FAILED`
`/home/dev/tao/project/metacat/build.xml:1732: /usr/share/maven2/boot/boot does not exist.`
` git status`
`On branch feature-autodetect-maven`
`Your branch is up-to-date with 'origin/feature-autodetect-maven'.`

`nothing to commit, working directory clean`

amoeba:
Yeah:

```sh
mecum@mn-sandbox-ucsb-1:~$ git clone https://github.com/nceas/metacat
fatal: destination path 'metacat' already exists and is not an empty directory.
mecum@mn-sandbox-ucsb-1:~$ mv metacat metacat-targz
mecum@mn-sandbox-ucsb-1:~$ git clone https://github.com/nceas/metacat
Cloning into 'metacat'...
remote: Enumerating objects: 26, done.
remote: Counting objects: 100% (26/26), done.
remote: Compressing objects: 100% (20/20), done.
remote: Total 24323 (delta 5), reused 14 (delta 4), pack-reused 24297
Receiving objects: 100% (24323/24323), 138.82 MiB | 64.09 MiB/s, done.
Resolving deltas: 100% (13611/13611), done.
Checking connectivity... done.
mecum@mn-sandbox-ucsb-1:~$ cd metacat
mecum@mn-sandbox-ucsb-1:~/metacat$ git fetch --all
Fetching origin
mecum@mn-sandbox-ucsb-1:~/metacat$ git checkout origin/feature-autodetect-maven
Note: checking out 'origin/feature-autodetect-maven'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at 8c035cb... Add support for deblike+maven2 to build.xml's setMavenHome target
mecum@mn-sandbox-ucsb-1:~/metacat$ git l^C
mecum@mn-sandbox-ucsb-1:~/metacat$ ant setMavenHome
Buildfile: /home/mecum/metacat/build.xml

setMavenHome:
     [exec] Result: 1
     [echo] Setting Maven home (maven.home) to '/usr/share/maven2/boot'.

BUILD SUCCESSFUL
Total time: 0 seconds
```

taojing2002:
So my error message shows:
`/home/dev/tao/project/metacat/build.xml:1732: /usr/share/maven2/boot/boot does not exist`
You see, there are two `boot` at the end of the file path. I guess somehow the maven home shouldn't have the `boot` at the end of file path.

amoeba:
I can run your script fine:

```sh
mecum@mn-sandbox-ucsb-1:/home/dev/tao/project/metacat$ ant setMavenHome
Buildfile: /home/dev/tao/project/metacat/build.xml

setMavenHome:
     [exec] Result: 1
     [echo] Setting Maven home (maven.home) to '/usr/share/maven2/boot'.

BUILD SUCCESSFUL
Total time: 0 seconds
```

How are you running that?

taojing2002:
The command `ant setMavenHome` works to me.
But the command `ant compile` fails. May you try this command as well?

amoeba:
Ah, I see. It's gotta be something with maven2. I'll look into it.

amoeba:
Alright, can you give it a whirl now?

taojing2002:
Works great! I will do some test, then merge it to dev-2.10. Thank you, 
Bryce!

amoeba:
Thanks for the help testing it out!

taojing2002:
You are welcome. Thank you for fix the issue!

Issue #1263, opened 2018-07-10
Update Metacat's sitemap implementation


amoeba:
To support features we're adding to [MetacatUI](https://github.com/NCEAS/metacatui), we need to take another look at the Sitemap implementation in Metacat. In particular, we want to support Member Nodes running Metacat w/ MetacatUI, and also the DataONE Coordinating Nodes running Metacat w/ MetacatUI as well. Together with the features we're adding to MetacatUI:

- [Injecting Schema.org/P418 JSON-LD in MetacatUI views](https://github.com/NCEAS/metacatui/issues/482)
- [Refactoring the MetacatUI router to not user fragment URLs](https://github.com/NCEAS/metacatui/issues/526)

Updating the Sitemap implementation will allow Metacat users running MetacatUI to be indexed by Google which would be awesome.

My plan is to:

- [x] Evaluate the current state of the Sitemap implementation
- [x] Make any updates needed to most effectively achieve the goal of getting Metacat instances indexable by Google
- [x] Test this implementation at appropriate scales to ensure it'll work both for Member Nodes and Coordinating Nodes

We have an old ticket, https://github.com/NCEAS/metacat/issues/563, that I think I'll close as this ticket obviates that one.

amoeba:
I see two big challenges:

- How will we tell Metacat about the URL format it should put in the sitemap file. Reducing the dependencies or complexity of dependencies between Metacat and MetacatUI is probably best. I don't think Metacat knows whether it's deployed with MetacatUI so it seems like an optional property with a sensible/valid fallback value makes a lot of sense
- The current implementation,

https://github.com/NCEAS/metacat/blob/7c255728da43aae0e9382dc82b39ba9d77fbeb62/src/edu/ucsb/nceas/metacat/Sitemap.java#L120

is too narrow a limit since not all Member Nodes only use EML. I think a better limit here is that the formatID must be one of the registered formats in the XML catalog.

mbjones:
@amoeba Did you note that metacat.properties already has the `guid.ezid.uritemplate.metadata` property, which is used to construct the canonical URL to be registered for DOI redirection?

Currently, the default is:

```
guid.ezid.uritemplate.metadata=/metacatui/#view/<IDENTIFIER>
```

[...]

PR #1283, opened 2018-10-03
The next generation of Metacat sitemaps

amoeba:
This is a minor code change but major refactor of how sitemaps work in Metacat.

The previous implementation of sitemap was intended to work with our old usage of Metacat, which was that Metacat serves its own content. Nowadays, our deployments use MetacatUI as a front-end to Metacat so the implementation needed to change.

What did I change?

- The SQL query that fetches records to go into the sitemap:
  - Now supports all metadata formats (by querying the `xml_catalog` table)
  - No longer includes archived content
  - No longer includes obsoleted content
- The filenames for the sitemap index and sitemap files themselves has changed to `sitemap_index.xml` and `sitemap{#}.xml`, respectively, to align with the common usage I saw on the web
- Configuration of sitemaps is quite different now. Because MetacatUI is now serving the content in Metacat, the sitemap URLs need to be from MetacatUI's URL space and not Metacat's as they are now. The two properties (`sitemap.location.base` and `sitemap.entry.base`) could be simplified a bit but I think what I went with is the most obvious. I could've also used the ezid configuration to control the URL space but would rather have clear configuration parameters for the two features rather than an arcane interaction between them.
- Sitemaps are now property XML escaped and URL encoded where appropriate
- Expanded the tests a bit to provide additional regression checking. I didn't add a regression test for the `archived = FALSE` part of the SQL query because my test setup is so broken but this'd be nice to do in the future
- User docs are updated to match implementation changes
- The `sitemap.excludedata` property was removed because it didn't appear to be used anyway

Closes issue: #1263 

Specific things I'd appreciate a review on is the SQL query (both its content and its potential performance implications) and the configuration properties.

### DIFF
--- a/docs/user/metacat/source/sitemaps.rst
+++ b/docs/user/metacat/source/sitemaps.rst
@@ -6,10 +6,20 @@ discussed in this section - which URLs on your websites are available for
 crawling. Currently, the only way for a search engine to crawl and index 
 Metacat so that individual metadata entries are available via Web searches 
 is with a sitemap. Metacat automatically creates sitemaps for all public 
-documents in the repository. However, you must register the sitemaps with 
-the search engine before it will take effect.
+documents in the repository that meet these criteria:
+
+- Is publicly readable
+- Is metadata
+- Is the newest version in a version chain
+
+However, you must register the sitemaps with the search engine before it will 
+take effect.
 
 
+Configuration
+-------------
+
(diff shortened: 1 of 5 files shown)


amoeba:
Yea, the subquery irked me a bit too but maybe Postgres does Smart Things. Thanks for looking at this!

PR #1301, opened 2018-11-28
DBTransform refactor for performance

nahf:
I refactored Transformer creation to be more efficient through the use of stylesheet-specific Templates.  Templates store the compiled stylesheet and are used to cheaply create new Transformers.   The net effect is only compiling each stylesheet once, and maintaining those in a static Map in DBTransform.  As there are not many templates, increased memory needed to store the map are considered to negligible. 

 I also created a new test class DBTransformTest to verify expected behavior and performance.

### DIFF
--- a/src/edu/ucsb/nceas/metacat/DBTransform.java
+++ b/src/edu/ucsb/nceas/metacat/DBTransform.java
@@ -36,12 +36,16 @@ import java.io.StringReader;
 import java.io.Writer;
 import java.sql.SQLException;
 import java.util.Enumeration;
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


taojing2002:
It should be merged to dev-2.10 first. So I close it.

mbjones:
Reopening this so that we get a clear review of the changes documented here from both Jing and Chris.  I want to be sure there are multiple Metacat devs that examine the impacts and benefits of this change.

csjx:
Closing this in favor of reviewing it under https://github.com/NCEAS/metacat/pull/1303

PR #1307, opened 2018-12-06
Feature d1 authorization optimization

nahf:
This feature-branch refactors authorization in edu.ucsb.nceas.metacat.dataone package to optimize the order of authorization tests from most efficient to least efficient (those needing to make API calls, for example).  Further work is possible here, including caching the Nodelist from the CN, but was seen as premature for now, as it would increase complexity, and the amount of improvement was uncertain.  Needs live testing in non-production environment.


nahf:
I switched jenkins to build dev / unstable from my forked repo (nahf/metacat), deployed to cn-dev UNM and profiled getSystemMetadata side by side with cn-dev-ORC.

Results showed 4-5 fold reduction in total time for a client to call getSysmeta on 100 different pids, and 50-100x reduction of time in authorization.

All objects tested were publicly readable objects, (which comprise about 85% of the collection), so performance increases will not necessarily be universal.

tabular results:
https://docs.google.com/spreadsheets/d/1FCZaABnuO214cfSe8GWXX7mz9S5LfobjaSfqfEAMESo/edit?usp=sharing

taojing2002:
Rob will create a new pull request on the branch feature-d1-authorization-optimization.

PR #1330, opened 2019-03-06
Installation file move into docs/user/metacat/source

yvanlebras:
If I'm not wrong ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ data are well and consistently described.
 - Owen Jones (owen.jones@imperial.ac.uk)
 
 ## Installation
-See the file "docs/install.html" for detailed instructions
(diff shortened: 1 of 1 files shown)


yvanlebras:
Thank you for your feedback. It seems to me there is no docs/install.html ? That's why I was searching towards src... I'm wrong?

taojing2002:
The "doc/install.html" references a relative html path of a installed 
Metacat. See you can find it at:
https://knb.ecoinformatics.org/knb/docs/install.html

yvanlebras:
Ooops. Really sorry... I understand now. Thank you!

taojing2002:
You are welcome. Thank you every much for your effort to improve Metacat!

PR #1335, opened 2019-03-13
Merge Dev 2.10 again for the 2.10.2

taojing2002:


csjx:
Also, @taojing2002 - What do you think about merging in the proposed changes in PRs https://github.com/NCEAS/metacat/pull/1330, https://github.com/NCEAS/metacat/pull/1331, and https://github.com/NCEAS/metacat/pull/1332? That look quite minor to me.

taojing2002:
Hi Chris:

The #1332 (about the command to create the metacat db user ) looks good 
to me. I tested it and the command worked on postgresql 9.3 as well. I 
am going to approve it.

The #1331 (about change from openjdk 7 to 8), since we still support 
java 1.7, i am not sure if we need to change it. What do you think?

The #1330 (about the change of the doc path). The change is wrong. The 
proposed the path is the source code path, not the our generated html 
page path. I will reject it.

Thanks,

Jing

Issue #1366, opened 2019-07-01
Publishing dataset with prov loses all prov relationships


laurenwalker:
Related to #1363 

**Steps to reproduce:**
1. Create a new dataset
2. Add provenance relationships in MetacatUI and click Save
3. After saving, click Edit, and edit the metadata in MetacatUI (e.g. change the title and click Submit)
4. Click the "view dataset" link to view the dataset landing page
5. Click "Publish with DOI"
6. Notice that the publish call succeeds, the page is refreshed, but all the provenance relationships are missing.

This happened in production to this dataset: 
Before: https://knb.ecoinformatics.org/view/urn:uuid:66e76292-46f9-4e51-8d54-d30ab87f4ec7
After: https://knb.ecoinformatics.org/view/doi:10.5063/F1KP80GX

And on dev.nceas:
Before: https://dev.nceas.ucsb.edu/view/urn:uuid:1e85a8ac-98f3-418d-9bd1-133c31c65039
After: https://dev.nceas.ucsb.edu/view/doi:10.5072/FK2445RW5D

gothub:
Regarding the refactor of metacat to parse all the DataONE ProvONE relationships, here some examples of current PROV and ProvONE and related terms and the namespace definitions:
```
dcterms  =  "http://purl.org/dc/terms/"
rdf = "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
prov = "http://www.w3.org/ns/prov#"
provone = "http://purl.dataone.org/provone/2015/01/15/ontology#"

dcterms:identifier
dcterms:Agent
rdf:type
prov:qualifiedAssociation
prov:wasDerivedFrom
prov:wasInformedBy
prov:hadPlan
prov:used
prov:wasGeneratedBy
prov:Association          
prov:WasAssociatedWith 
prov:Agent                
provone:Program
provone:Execution
provone:Data
provone:User
```
This set of terms does not contain all possible ProvONE terms, but just the ones that are currently implemented in the R client and MetacatUI.

Compiling an exhaustive list of all potential PROV and ProvONE RDF relationships, and then querying a parsed resmap to determine which relationships to carry forward, seems to be an error prone method and precludes future relationships that may be added in the future.

One approach would be to retain any relationships that use the PROV and ProvONE and RDF namespaces.

mbjones:
[...]

PR #1367, opened 2019-07-12
Dev 2.10 for 2.10.4 release

taojing2002:


src/edu/ucsb/nceas/metacat/dataone/resourcemap/ResourceMapModifier.java
```diff
@@ -0,0 +1,446 @@
...
+     * @param obsoletedBys  a map represents the ids' with the obsoletedBy relationship - the keys are the one need to be obsoleted (replaced); value are the new ones need to be used. They are all science metadata objects
+     * @param newResourceMap  the place where the created new resource map will be written
+     * @param subject  the subject who generates the resource map
+     * @throws UnsupportedEncodingException 
+     */
+    public void replaceObsoletedIds(Map<Identifier, Identifier>obsoletedBys,  OutputStream newResourceMap, Subject subject) throws UnsupportedEncodingException {
+        //replace ids
+        Vector<String> oldURIs = new Vector<String>(); //those uris (resource) shouldn't be aggregated into the new ore since they are obsoleted
+        Vector<String> newURIs = new Vector<String>(); //those uris (resource) should be added into the new aggregation
+        if(obsoletedBys != null) {
```
csjx: Minor formatting for readability: add spaces in flow control statements - change `if(obsoletedBys != null)` to `if (obsoletedBys != null)`, etc.

src/edu/ucsb/nceas/metacat/dataone/resourcemap/ResourceMapModifier.java
```diff
@@ -0,0 +1,446 @@
...
+     */
+    private void replaceAggregations(Model model, Resource oldOREResource, Resource newOREResource, Vector<String> oldURIs, Vector<String> newURIs) {
+        //create a aggregation resource for the new ore id
+        Resource newAggregation = ResourceFactory.createResource(newOREResource.getURI() + AGGREGATION);
+        Property predicate = ResourceFactory.createProperty(ORE_TER_NAMESPACE, "isDescribedBy");
+        Statement statement = ResourceFactory.createStatement(newAggregation, predicate, newOREResource);
+        model.add(statement);
+        
+        Vector<Statement> needToRemove = new Vector<Statement>();
+        Resource oldOreAggregation = model.getResource(oldOREResource.getURI()+AGGREGATION);
```
csjx: Minor formatting: add spaces between operators.

src/edu/ucsb/nceas/metacat/dataone/resourcemap/ResourceMapModifier.java
```diff
@@ -0,0 +1,446 @@
...
+import com.hp.hpl.jena.rdf.model.Selector;
+import com.hp.hpl.jena.rdf.model.SimpleSelector;
+import com.hp.hpl.jena.rdf.model.Statement;
+import com.hp.hpl.jena.rdf.model.StmtIterator;
+
+import edu.ucsb.nceas.metacat.properties.PropertyService;
+
+/**
+ * This class will create a new resource map by modifying a given resourceMap input stream. 
+ * @author tao
```
csjx: @taojing2002 - This modifier class looks great.  I reviewed the logic, and it looks like you are changing all of the appropriate statements (`documents`/`isDocumentedBy`, `aggregates`/`isAggregatedBy`, `describes`/`isDescribedBy`).  The one question I have is about the `dcterms:identifier` statements.  Can you confirm that the old obsoleted resources are removed, which also removes the statements that include the old `dcterms:identifier` literal, like `terms:identifier>urn:uuid:326e21d5-c961-46ed-a85c-28eeedd980de</terms:identifier>`? I didn't see that in the code, by might have missed it.

src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
```diff
@@ -2209,17 +2212,23 @@ public Identifier publish(Session session, Identifier originalIdentifier) throws
 				oreSysMeta.setObsoletedBy(null);
 				oreSysMeta.setSize(BigInteger.valueOf(resourceMapString.getBytes("UTF-8").length));
 				oreSysMeta.setChecksum(ChecksumUtil.checksum(resourceMapString.getBytes("UTF-8"), oreSysMeta.getChecksum().getAlgorithm()));
+				oreSysMeta.setFileName("resourceMap_" + newOreIdentifier.getValue() + ".rdf.xml");
 				
 				// ensure ORE is publicly readable
-				oreSysMeta = makePublicIfNot(oreSysMeta, potentialOreIdentifier);
-				
+                oreSysMeta = makePublicIfNot(oreSysMeta, potentialOreIdentifier);
```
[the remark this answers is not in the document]
taojing2002: The new system metadata of the new resource map object was copied and modified from the old resource map system metadata. Since the old resource map is not necessary to be public readable, we do a check here.

src/edu/ucsb/nceas/metacat/dataone/resourcemap/ResourceMapModifier.java
taojing2002: I hesitate to remove those old obsoleted resources since they may be on some triples we don't understander. Otherwise those triple may point to an orphan.  
Based on the strategy from this ticket - https://github.com/NCEAS/metacat/issues/1366, we only modify those triples we understand ( `documents/isDocumentedBy`, `aggregates/isAggregatedBy`) and reserve the all others. So I also keep those obsoleted resource in case they exist on some triples.

csjx: I see what you mean @taojing2002 - That certainly is an issue, but I think we also need to remove all unused triples so they don't accumulate and make the resource map harder to interpret or even validate.  So I think we might need to query the Jena model and check for `subject` or `object` instances with the `identifier` literal or its `URI` equivalents (URL encoded or not).  If the identifier is used in new triples, then yes, we keep them. But if not, we remove them.  We probably should also try to validate the resulting ORE by re-parsing it with Jena and see if it throws any errors due to orphan nodes.  Let's discuss this with @gothub and @laurenwalker since they've both worked on RDF mofification in MetacatUI - we do the same operations there too.

Issue #1369, opened 2019-07-17
Refactor sitemap generation approach to be XML-aware


amoeba:
I see 

```xml
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:sm="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd">
<sitemap>
<loc> https://neutral-cat.nceas.ucsb.edu/sitemap1.xml </loc>
<lastmod>2019-07-16</lastmod>
</sitemap>
</sitemapindex>
```

and, from looking at the code, it looks like our attempt at making pretty-printed XML with string concatenation isn't working so well. Should be a quick fix.

mbjones:
Are we using string concatenation to create XML?  I think that's a bad idea, and will likely produce XML with well-formedness problems when (not if) we encounter escaping situations we didn't anticipate.  Wouldn't it be reasonably easy to stream out through an XML parser rather than using string maninuplation?

amoeba:
I think changing the implementation is reasonable. It's been strings-only for a long time AFAICT and I've never seen issues. When I went back in to the code last year I did add some additional XML-escaping to make it safer but I agree about the increased risk for escaping. This doesn't sound like a big task so I'll take a look.

amoeba:
Alright, this is all done and working as expected. I refactored everything that generate XML strings to use the `DocumentBuilderFactory/DocumentBuilder/Document/etc` gang and tested how it handles URLs with PIDs with special characters and confirmed the underlying functionality of actually generating sitemaps works as expected.

I'll leave open and let the commit auto-close this when merged into master.

amoeba:
Heya @taojing2002 , I've [PR'd](https://github.com/NCEAS/metacat/pull/1393) this and could use a second pair of eyes. Do you think you could target this for the next Metacat release?

Issue #1368, opened 2019-07-16
Reconfigure how Metacat runs the sitemap task


amoeba:
Currently, sitemaps are implemented in a `Sitemap` class which inherits from `java.util.TimerTask` and handles the responsibility of querying the database for metadata records and writing sitemap XML files to the `./sitemaps` directory. Sitemaps are configured by default but not enabled.

When `MetacatServlet` inits, amongst many other things, it also inits `MetacatHandler` with a `java.util.Timer` which is later on used to schedule the Sitemap task. But at this point, the Sitemap task isn't scheduled to run.

The only way to get sitemaps to begin generating is to trigger `MetacatServerlet#handleGetOrPost` by sending an HTTP request to `/metacat/metacat` with any parameters (i.e., sitemaps are forcibly scheduled regardless of what `action` equals). At this point, `MetacatHandler`'s `java.util.Timer` instance is asked to schedule the Sitemap task unless it already is:

```java
timer.schedule(smap, firstDelay, sitemapInterval); // smap is a Sitemap instance
```

Is there a reason we had Sitemaps set up this way? I'd like to see this changes so that sitemaps, if configured, are on by default. We could add an admin API command, `regenerateSitemaps`, to force sitemaps to regenerate.

mbjones:
Your changes sound good to me.  A config parameter with sitemaps on or off would be best, and Metacat should detect that param change while running (i.e., not require a Metacat or tomcat restart to pick it up).

amoeba:
Great, I'll look into that.

amoeba:
Heya @taojing2002 , I've [PR'd](https://github.com/NCEAS/metacat/pull/1393) this and could use a second pair of eyes. Do you think you could target this for the next Metacat release?

PR #1393, opened 2019-10-03
Improve sitemap functionality

amoeba:
This PR addresses some outstanding issues with sitemaps not addressed in our last pass over the implementation and makes some minor improvements.

Changes:

- Refactors how sitemaps get created. Previously, the sitemap task was triggered only directly and they the sitemap task is run on startup (if enabled). ref #1368 
- Sitemaps XML files are now created in an XML-aware way instead of with string manipulation ref #1369 
- Sitemaps are enabled/disabled via the `sitemap.enabled` property in metacat.properties now, rater than implicitly being on
- The Metacat admin panel has a new sitemaps section which controls all of the sitemap config so the operator doesn't need to dive into metacat.properties
- Documentation is updated to reflect all of this

@taojing2002 could you have a look? Specifically, did I understand how to use the `DocumentBuilderFactory`, `DocumentBuilder`, `Document` classes in a clean way?


amoeba:
During testing, @taojing2002 saw an error:

```
Exception in thread "Timer-0" java.lang.NullPointerException
	at edu.ucsb.nceas.metacat.Sitemap.generateSitemaps(Sitemap.java:306)
	at edu.ucsb.nceas.metacat.Sitemap.run(Sitemap.java:145)
	at java.util.TimerThread.mainLoop(Timer.java:555)
	at java.util.TimerThread.run(Timer.java:505)
```

No idea what this is yet as the stack trace isn't overly helpful. We'll hold off on merging this PR until this is fixed.

amoeba:
Okay @taojing2002 the NPE is fixed. I failed to guard access to a variable with a null-check. The reason you found this was that you had a totally fresh Metacat install. After you add an object or two, the NPE goes away. I musta not seen it because I was installing Metacat fresh and then loaded up objects and wasn't seeing the NPE in the logs.

Can you give this a whirl tomorrow?

PR #1414, opened 2019-12-12
[WIP] New Export Format

ThomasThelen:
## New Export Format

This PR introduces changes, primarily in `MNodeService.getPackage`, to update the package export to the new version.

### To Do:

1. Revise File handling (I need to pick a directory for temp files & look at efficiency)
2. Test more
3. Unit tests
4. Developer Documentation

#### getPackage
1. I moved the validation code that was at the beginning out into a separate method
1. Removed filename parsing from the EML (as per a call with chris)
1. I created all of the directories (metadata, data, sysmeta, etc) towards the top of the method because these directories are written to throughout the method. I thought that defining them all together, rather than as they're used is helpful for reading.

#### Misc
1. The changes made aren't _too_ extensive, and can be thought of as 6 parts.
   a. Generating JSON-LD
   b. Generating HTML file table
   c. Generation of README.HTML (Contains a&b)
   d. Creating the new bag directories (metadata, sysmeta, etc)
   c. Searching for prov:atLocation
   d. Adding files to d, optionally using c

### README.html
The README.html is a new file that is added to the bag root. It includes a minimal JSON-LD representation of the EML in the head. It also includes a table which lists all of the data objects.

##### Misc Notes
1. The vanilla readme will _only_ be present in the bag if the JSON-LD and file table generation fails. 

##### MNodeService.generateReadmeHTMLFile
The README.html is generated in`MNodeService.generateReadmeHTMLFile`, which gets called by `MNodeService.getPackage` after the Identifier of the EML doc is found. 

Within this method, the file table and JSON-LD are generated and inserted into the README. Writing the bytes to this file is also done in `generateReadmeHTMLFile`.

Since the XSLT requires the EML, we require (and pass in) the user's session. The file-table generation requires the system metadata files for each data object. We don't want to add the EML or ORE, so we also need knowledge of these PIDS. The function parameters `List<Identifier> metadataIds, List<Identifier> coreMetadataIdentifiers` take care of these, respectively.

##### MNodeService.generateReadmeBody
When `MNodeService.generateReadmeHTMLFile` is ready to generate the HTML representation of the EML document, this method is called. It handles the XSLT, which is fairly straight forward.

##### MNodeService.createSysMetaTable
This method handles the creation of the HTML table that is eventually inserted into the README. It loops over each metadata identifier, and for each, makes sure that it's not describing an ORE or EML and then passes it to the XML transformer. Is uses the XSL at `lib/style/skins/package-export/package-export.xsl`. 

Note that the XSL _only _ creates a row in a table, not the actual table.
At the end of the method, we need to wrap the rows in an HTML table, which is hackily done.

##### MNodeService.insertTableIntoReadme
Combines an HTML table and a readme document. It searches for a hardcoded div in the vanilla readme body and places the table inside. Returns the vanilla readme if something went wrong.

##### Misc Notes
1. The JSON-LD is created by using the `Transformer` class to use the xsl at `lib/style/common/conversions/emltojson-ld.xsl` with the EML document.

2. Insertion into the README is hacky, notes on it in the `MNodeService.insertJSONIntoReadme` section.

3. It happens after the file table stuff

##### MNodeService.generatePackageJSONLD
To generate the JSONLD, we need to have a few things.
1. The EML as a String
2. The path to the EML->JSON-LD XSL file

This method obtains both of those in the normal ways (done similarly in other places in the code base)
These are then passed to `MNodeService.emlToJson`.

##### MNodeService.emlToJson
This method performs the actual transform. As with the two methods above, it returns an empty string if something went wrong.

##### MNodeService.insertJSONIntoReadme
After the vanilla readme was created and the JSON-LD created, this method is called to insert the JSON-LD into the README. It searches the readme string for a `<head>` tag, and inserts it within.


taojing2002:
Close the obsoleted request.

PR #1433, opened 2020-03-12
Dev 2.12 for the systemmetadata query performance

taojing2002:


taojing2002:
Hi Chris, the first commit  is in the `HazelcastService`. Do you think we need to change more?
Thanks!

PR #1486, opened 2021-02-23
Replace the 0.97 BagIt Export with SpeedBagIt

ThomasThelen:
This PR replaces the current export format code with SpeedBagIt, hopefully providing better performance for downloaded datasets.

#### MNodeService.getPackage
The first changes you'll see are the replacement of the `BagFactory` and `Bag` classes with a`SpeedBagIt` class using the BagIt 0.97 format.
Since we down't have any temporary files in this implementation, the list holding the temporary files was also removed.
Due to the way that HtmlToPdf works, we _have_ to write some of those artifacts to disk. I didn't modify this code but added some comments around it. If we _really really_ wanted to avoid filesystem writes we could modify HtmlToPdf to write to a stream instead of disk.
Since SpeedBagIt works with streams, and we don't want to leave any filesystem artifacts around, we load the pdf file into memory and delete the file along with the unwanted artifacts (the html & css files).

Next, the data files in the data package are processed. Most of this code was reused from the current implementation only real addition is that we get a stream to each file and add it to the SpeedBagIt object rather than writing them to disk in a temporary location. The last step is adding a stream to the pid-mapping.txt file and declaring it as a tag file.

#### MNResourceHandler.getPackage (V1 and V2)
The only changes here were related to the naming of the package, specifically _where_ the package is named. In the current implementation, the name of the package is set in the MNodeService call, but the MNResourceHandler function that calls it needs to know about it. BUT. It also needs an InputStream from MNodeService. The previous developer had two options, either return a tuple from MNodeService as (filename, bagStream) or stuff it in the DeleteOnCloseFileInputStream class. Right now it's stuffed in the DeleteOnCloseFileInputStream class-which we no longer need because there's no file to delete when the stream finishes. I think the real reason it was in the node service was because the LoC BagIt library needed to create a zip file, so it made sense to name it there. I deleted that class since there's no need for it anymore.
Rather than make things confusing with returning a tuple, I just cut+pasted the naming rules out of MNodeService and put them here instead. There's already code for naming the package (that I don't think would ever get hit) here so I didn't see any issue with it.

### WIP Tag
- [x] Rebase with develop


ThomasThelen:
We're going to merge/review this after the .15 release

ThomasThelen:
@taojing2002 has reviewed and blessed this for merging.

PR #1513, opened 2021-06-25
Develop 2.15.0 to main (second time)

taojing2002:
Issue #1499 Removed the classes FilterCommonRootSolrField, FilterGroupProcessor, FilterRootElement, FilterProcessor from metacat-index, as these classes are now included in DataONE/d1_cn_index_processor. Also updated the Spring context file that accesses these classes.


csjx (approved):
Thanks @taojing2002 - Looks good to me.

Issue #1517, opened 2021-07-27
Update EML Semantic Annotation indexing to include and expand property URIs


amoeba:
**Note: This issue is duplicated from https://github.com/DataONEorg/d1_cn_index_processor/issues/27**

I'll track work over on https://github.com/DataONEorg/d1_cn_index_processor/issues/27 but will keep this ticket up to date with respect to its status and the checkable items below.

---

EML Semantic Annotations are represented in EML using a structure like

```xml
<annotation>
  <propertyURI label="some label">http://example.com/some_uri</propertyURI>
  <valueURI label="some other label">http://example.com/some_other_uri</valueURI>
</annotation>
```

To provide search for the above metadata, we extract and parse the character data from the `valueURI` element as an IRI, query the OntologyModelService for any parent classes for the IRI, and smush all these terms together in the `sem_annotation` field. We kept the indexing rules narrowly-focused as a start because we were planning on using EML Semantic Annotations narrowly to start with. It's catching on within our teams and also within external teams and the use is outstripping the implementation.

Over on https://github.com/NCEAS/metacatui/issues/1807, I'm breaking apart the popover widgets we show on dataset landing pages that contain EML Semantic Annotations into two separate popovers: One for the `propertyURI` and one for the `valueURI`. A key part of that widget is a link that searches for other datasets annotated with the term you're viewing. Because `propertyURI`s aren't being expanded and stored in the search index, searches for datasets annotated with a specific `propertyURI` don't work.

I propose we expand what we store in the `sem_annotation` field to cover the the `valueURI` _and_ `propertyURI` and of course any expanded terms (superclasses for `valueURI` and superproperties for `propertyURI`). I could see us developing a more structured indexing approach for EML Semantic Annotations but I don't think we need it at this point so I'm opting for the small change.

This change will will require re-indexing the ~200-300 EML docs with semantic annotations in them. The number might grow before re-indexing is complete.

- [x] Make the changes in this codebase and metacat-index where the logic is duplicated
- [ ] Deploy changes to CN and MN stacks
- [ ] Re-index affected content

amoeba:
PR'd and merged onto develop.

PR #1518, opened 2021-07-28
Update EML annotation support

amoeba:
Implements https://github.com/NCEAS/metacat/issues/1517

1. Updates eml-semantics XSL to support individual popovers for the property and the value (in MetacatUI)
2. Updates the indexing code to now also include property URIs (and their super-properties) in the index


amoeba:
Hey @taojing2002, do you want to take a look at the changes here before one of us merges? The changes are backed by updated tests and I've also tested with a local deployment but a second set of eyes would be great. I threw this under the 2.16.0 milestone but whatever the next release is would be great.

taojing2002:
Hey @amoeba . It looks good to me. You may merge it at your convenience. Thanks.

PR #1572, opened 2022-03-13
Ignore sanitizing the resource map filename (it's always the same) an…

ThomasThelen:
Hey Jing,

This PR fixes the bug where an extra `xml` was being appended to the science metadata filename which was happening [here](https://github.com/NCEAS/metacat/compare/develop...feature-hierarchical-download?expand=1#diff-eb84e3003f3b0cfc13db6c6e4e4bdcf186a1cafc847c0a1b7d6e53cca6662c63L240). Because the filename is completely controlled by metacat, I removed that sanitation call to it (the filename will always be science-metadata.xm).

I also looked into the issue where some data packages on dev.nceas had system metadata filenames with slashes in place of `:`. I made a small change to the sysmeta naming routine, tested it on the demo server, and wasn't able to reproduce it. I put in two debug log statements incase it crops up again on dev so that I can debug it a little better.

### DIFF
--- a/src/edu/ucsb/nceas/metacat/download/PackageDownloaderV2.java
+++ b/src/edu/ucsb/nceas/metacat/download/PackageDownloaderV2.java
@@ -67,8 +67,10 @@ import com.hp.hpl.jena.rdf.model.Model;
 
 
 /**
- * A class that handles downloading data packages under the V2 format.
- *
+ * A class that handles downloading data packages under the V2 format. It contains a number of private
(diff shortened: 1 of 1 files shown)


taojing2002:
Hi Tommy:

Thank you! So have you fixed all known bugs? Is it ready to build another
beta release?

Regards,

Jing

ThomasThelen:
AFAIK this has all of the bugs fixed, and is ready for another beta.

taojing2002:
Cool. Thank you, Tommy.

Jing

taojing2002 (approved):

PR #1613, opened 2023-03-22
Feature 1612 ant streamlining

artntek:
1. Cleaned up some old commented-out stuff > 8 years old (c76a3ba1662d0a0814410e6e634fddf85ee1efd1), and reformatted in line with google coding guidelines (65cf0a31179908f25909e2511dbda3b6966477ec - mainly tabs to spaces). Pushed these changes separately, so they didn't mess up diff for the real changes

2. Changed the `install` target so it no longer builds or installs the documentation (javadoc or admin docs), This saves time (and reduced the ant wall of text) when developing.

3. Added a new `full-install` target that does install the docs (it has exactly the same functionality as the old `install` target)

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -811,12 +811,12 @@
         </java>
     </target>
 
-    <target name="install" depends="warNoDocs, dist-metacat-index, build-metacat-ui,deploy"
+    <target name="install-dev" depends="warNoDocs, dist-metacat-index, build-metacat-ui,deploy"
             description="* * * Install Metacat For Development. Lightweight install: does not include docs * * *">
         <echo>Dev install completed at ${NOW}</echo>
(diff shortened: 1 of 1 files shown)


mbjones:
Nice. We have some CI build systems that are configured to run `ant` -- mainly our jenkins process in DataONE. What do you think about naming the full install process back to `install`, and instead create a new target called `install-dev` that does the more limited build. I'm not sure where we might have referenced that current target name, but seems safer to make the new target have the new behavior. Did you update `fulldist` target as well? I think we reference these targets in Jenkins builds at a minimum.

taojing2002 (approved):
It looks good. Thank you, Matthew!.

Issue #1443, opened 2020-06-01
getPackage implementation doesn't handle duplicate filenames


amoeba:
While doing testing on https://github.com/NCEAS/metacatui/issues/1397, I noticed some inconsistent behavior with the `getPackage` implementation: If two objects in a package share the same `fileName` property in sysmeta, you don't see both objects in the `data` folder or in the `manifest-{algo}.txt` file. But you do see both in the `pid-mapping.txt` file. A quick glance at the code backs this up.

AFAIK there's nothing codified in the DataONE API as to what the expected behavior is. This is definitely an edge case but I think we can at least do something like myfile, myfile (1), etc or myfile-{pid1}, myfile-{pid2}, etc when we have duplicates.

mbjones:
Thanks @amoeba for the report. This is a special case of the larger problem of file naming consistency in user-supplied data packages.  We also need to be watching out for cases where the filename is not legal on a given platform, or even is maliciously crafted to exploit security bugs. So, all package naming in both our `get()` and `getPackage()` operations needs to use a filename mangling routine that deals with these use cases consistently. At a minimum, deal with:

1) filename uses illegal characters
2) filename is too long for platform
3) filename is not unique
4) filename is/isn't unique on case-insensitive filesystems
5) filename has malicious payload designed to exploit weaknesses

Probably others I am not thinking of.  At least some of these I think are already handled in `get()` when setting Content-Disposition headers I think.

ThomasThelen:
I can definitely handle this case in SpeedBagIt, which currently throws `java.util.zip.ZipException: duplicate entry` when two files are written to the same location. This behavior is largely undefined when it comes to the BagIt standard and I don't think that the library can do much else.

### Solution 1: Don't fix it
Catch the exception from SpeedBagIt and re-throw to the user letting them know that they have a package that is skewed (they're making two conflicting claims about a file). The problem with this is that it's a hassle/burden on the user to have to fix the sysmeta _and_ issues arise if one of the files is used in another package where the path is perfectly valid in that context.

### Solution 2: Check the filenames before adding to SpeedBagIt
As @amoeba  mentioned we can rename the files to something that we think is sensible. Renaming the file will break consistency with the sysmeta document that's also stored in the bag which might be considered a bug, but at least the user is actually getting the data with slightly misleading metadata. This _may_ be an issue if bags are ever consumed by DataONE because there would be that mismatch of filenames on disk and filenames claimed by the sysmeta documents.

[...]

PR #1614, opened 2023-03-23
Inital refactor to use speedbagit 1.0.4 to handle duplicate filenames.

mbjones:
See issue #1443.  Depends on speed-bagit 1.0.3, which needs to be installed for testing.

Note I did not yet fully test this code, so it needs to be checked in an operational test environment with a test written with a package that has duplicated names for data files. Opening this draft PR so we can review more easily.


artntek:
Completed so far:
1. updated pom file to depend on speedbagit 1.0.4 instead of 1.0.3
2. verified this builds and runs
3. created a test dataset with two identically-named files, and verified that when downloaded, one has been renamed by pre-pending "0-duplicate-" as expected
4. discovered that we have no test coverage for `bagit-1.0` downloads, only for `bagit-097`.
5. discovered that we have no test coverage for `src/edu/ucsb/nceas/metacat/download/PackageDownloaderV2.java`; the existing V2 test actually tests V1

artntek:
Wrote up issues for stuff I noticed during this PR work 
1. https://github.com/NCEAS/metacat/issues/1616 
2. https://github.com/NCEAS/metacat/issues/1618

src/edu/ucsb/nceas/metacat/download/PackageDownloaderV2.java
```diff
@@ -208,17 +209,26 @@ public void addDataFile(SystemMetadata systemMetadata, InputStream inputStream)
...
-			this.speedBag.addFile(inputStream, dataPath, false);
-		} catch (Exception e) {
-			logMetacat.error("Error adding the datafile to the bag", e);
-			throw new ServiceFailure("There was an error creating the temporary download directories.", e.getMessage());
+			try {
+				this.speedBag.addFile(inputStream, dataPath, false);
+			} catch (SpeedBagException e) {
+				String duplicateDataPath = Paths.get(dataDirectory, "0-duplicates", dataPath).toString();
+				logMetacat.warn("Duplicate data filename, renaming file to add to bag: " + duplicateDataPath, e);
+				this.speedBag.addFile(inputStream, dataPath, false);
```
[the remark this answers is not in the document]
artntek: fixed in 7a55030e8dc12de967e28b18fdfa284c065335e0

mbjones:
I added another issue I noticed when doing a quick review of your commit:

3. #1617

artntek (commented):
finishing my review, since I submitted more fixes, so would like reviews from others :-)

src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
taojing2002, line 1: We should keep the copy right and GPL statement in the file

artntek: Had a discussion with @mbjones recently in #devteam thread, concluding we can omit copyright/license statement from newly-created files, and remove statements from older files as we touch them - see: https://nceas.slack.com/archives/C0N49FM9V/p1680621985374409?thread_ts=1680545835.114129&cid=C0N49FM9V

taojing2002:
Oh, I didn't notice that. Thanks.

artntek:
added a fix & test for Issue #1616

src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
taojing2002, line 1: Line 3063: `fileName = "0-duplicate-" + fileName;` adds a hard code prefix string to the duplicated file name. It works with the package with two identical file names. I guess this approach will fail if the package has more than two identical file names.

src/edu/ucsb/nceas/metacat/download/PackageDownloaderV2.java
taojing2002, line 1: Line 192: `String duplicateDataPath = Paths.get(dataDirectory, "0-duplicates", dataPath).toString();` adds a hard code prefix string for the duplicated file name. It works with the package having two identical file names. I guess this approach will fail if the package has more than two identical file names.

src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
mbjones: yeah, agreed. I thought about making it a loop until we got a name that works, but that would be more complicated, and I thought it would be rare if more than one rename would be needed (the likelihood of collision is much lower for the rename). If the rename fails, then the second exception means that the bag download will fail, but this time with a clear error and it won't just invisibly fail. So maybe the risk is small? Your call on whether to make it more of a loop.

taojing2002:
Yeah, adding a loop is complicated. So may we just append a 4 digits random
number in the file name? This can handle about 9999 duplicate file names.
Or safer, we can append current time + 2 digits random number.

artntek:
Jing and I discussed a simple solution - instead of using `fileName = "0-duplicate-" + fileName;`, we could simply define an `int index = 0;` and then replace the rename with:

`fileName = index++ + "-duplicate-" + fileName;`

This way, it will use a different integer every time something is renamed - so no duplicates are possible. the integers may not be contiguous (for example if we have 4 files that comprise 2 sets of 2 identical files each), but that isn't really an issue, afaict. I will make this change and resubmit

taojing2002 (approved):
Looks good!

PR #1621, opened 2023-04-27
Second merge for 2.19.0 from develop to main

taojing2002:


docs/user/metacat/source/install.rst
```diff
@@ -482,27 +482,27 @@ search engine. Unfortunately the Solr Debian packages that come with the Ubuntu
 system are obsoleted and you have to install the binary packages by yourself. This section 
 provides guidance on how to setup Solr to run in production on *nix platforms, such as Ubuntu.
 
-Metacat supports Solr 8.8.2 and newer versions. You may download the binary releases from:
+Metacat only supports ``Solr 8.*``, ``NOT 9.*``. You may download the binary releases from:
 
 https://lucene.apache.org/solr/downloads.html
```
artntek, line 487: suggestion - could use the direct link to v8, since we don't support v9? https://solr.apache.org/downloads.html#solr-8112

docs/user/metacat/source/install.rst
```diff
@@ -482,27 +482,27 @@ search engine. Unfortunately the Solr Debian packages that come with the Ubuntu
...
 
-1. Go to the directory which contains the Solr release file and extract the installation script file by typing (assume the download file being solr-8.8.2.tgz):
+1. Go to the directory which contains the Solr release file and extract the installation script file by typing (assume the download file being solr-8.11.2.tgz):
```
artntek, line 489: suggested reword:

"(**assuming** the download file **is** solr-8.11.2.tgz)

docs/user/metacat/source/install.rst
taojing2002: The problem using the direct link is that the link will be invalid if a newer version comes up. It only keeps the newest versions on that page.

artntek (approved):

Issue #1631, opened 2023-05-27
Abort Metacat Startup if Misconfigured and/or Critical Resources are Inaccessible


artntek:
## Acceptance Criteria
* Metacat startup is aborted if `metacat-site.properties` is not found AND `/var/metacat` is not writeable by the metacat user (details below).
* Logs contain clear messaging about what caused the aborted startup
* Logs contain clear information to help the reader fix the problem

========= 

### After the functionality is merged, iteratively add detection for other critical resources/services, e.g:
 *  SOLR is running, right version, port is accessible (at least for k8s - see note below)
 *  database is running and accessible (at least for k8s - see note below)
 *  /var/metacat exists and is writable by the web user
 *  properties file exists or directory is writable

*NOTE: be careful what we add here! Sometimes, we want to allow Metacat to start with
incomplete dependencies, since these are configured later through the admin interface --
 e.g. database connection & solr for non-k8s deployments*

============================================

## Details:
1. On startup, metacat tries to load the site properties from the default location (`/var/metacat/config/metacat-site.properties`)
2. If it doesn't exist, metacat tries to create a new, empty file at this same location. If it succeeds, all is well 
3. However, there’s an edge case where that doesn't work (because /var/ isn't writeable, for example). If this happens, there are clear errors in the logs:

(catalina.out): 
```
metacat 20230524-18:51:16: [ERROR]: PropertiesWrapper.initSitePropertiesFilePath(): 
a general I/O error occurred while trying to create directory hierarchy or actual 
file for site properties at: /var/metacat/config/metacat-site.properties; see 
exception message for details [edu.ucsb.nceas.metacat.properties.PropertiesWrapper:logAndThrow:482]
java.nio.file.AccessDeniedException: /var/metacat
(+ stack trace…)
```
(and localhost.<date>.log): 
```
javax.servlet.ServletException: Service problem while intializing MetaCat Servlet: 
PropertiesWrapper.initSitePropertiesFilePath(): a general I/O error occurred while 
trying to create directory hierarchy or actual file for site properties at: 
/var/metacat/config/metacat-site.properties; see exception message for details : /var/metacat
(+ stack trace…)
```
…but the metacat app continues to limp along and throw `NullPointerException`s whenever something makes a call to PropertyService.

also see [google doc](https://docs.google.com/document/d/1nm2f8dDKJvUKGQ2OQdIz_jldk0-jlYn_jfLD-o2n-JM/edit?disco=AAAAwEq9mkY)

[...]

PR #1633, opened 2023-05-31
Feature 1631 startupabort

artntek:
fix for #1631 -  Abort Metacat Startup if Misconfigured and/or Critical Resources are Inaccessible 

- Metacat startup is aborted if metacat-site.properties is not found AND /var/metacat is not writeable by the metacat user (details below).
 metacat log (typically tomcat/logs/catalina.out) and host log (tomcat/logs/hostname.(date).log) contain clear messaging about what caused the aborted startup and how to fix the problem


taojing2002 (approved):
Looks good to me. Thank you!

Issue #1635, opened 2023-06-05
Enable Passing Secret Credentials to Metacat via Environment Variables


artntek:
Related to Epic #1623  

This is a prerequisite for the Helm deployment work.

As an admin deploying metacat via helm, I want to be able to `kubectl apply` my `secrets.yaml` file and have those values be read from environment variables by metacat when it starts up, so that I don't have to hardcode those secrets into the properties file, which is insecure.

artntek:
fixed in PR #1636

PR #1636, opened 2023-06-06
Feature 1635 env secrets

artntek:
see Issue #1635 - "Enable Passing Secret Credentials to Metacat via Environment Variables"

For k8s deployments, secrets are kept in a kubernetes secrets object. In order to allow passing these secrets to Metacat during a helm deployment, without adding them to the properties files (plain text; insecure), this PR enables metacat to do the following:

1. Read a new property `application.envSecretKeys` from the properties file, which contains comma-delimited mappings between: camelCase / period-delimited properties keys (as found elsewhere in the props files) and SCREAMING_SNAKE_CASE environment variable keys.

2. upon receiving a call to getProperty(), checks if that particular property is in the mappings read in (1), above. If so, it tries to retrieve the secret it using System.getenv(). If it's not set, metacat then looks in the regular properties file for a fallback value (for backwards compatibility)


src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -281,6 +301,22 @@ protected Path getMainMetadataFilePath() {
         return mainMetadataFilePath;
     }
 
+    /**
```
artntek, line 304: FYI, I refactored this to here form PropertyService, since I call the same code here from within initialize(), so this makes it simpler and follows DRY

test/edu/ucsb/nceas/LeanTestUtils.java
artntek, line 1: these were just changes to add more debugging and validation to help track down other test failures

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -281,6 +301,25 @@ protected Path getMainMetadataFilePath() {
...
+    protected void syncToSettings() throws GeneralPropertyException {
+        try {
+            Settings.getConfiguration();
+            Settings.augmentConfiguration(this.getDefaultPropertiesFilePath().toString());
+            Settings.augmentConfiguration(this.getSitePropertiesFilePath().toString());
+        } catch (ConfigurationException e) {
+            logAndThrow("PropertiesWrapper.syncToSettings(): augmenting DataONE settings: ", e);
+        }
+    }
+
```
taojing2002, line 322: To my understanding, those environmental variables wouldn't be stored in the site properties file. So it seems their values will not be propagated to settings. Do we need code to set the Settings for those environmental variables?

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -409,6 +443,50 @@ private void initialize() throws GeneralPropertyException {
...
+        final String ENTRY_DELIMITER = ":";
+        final String KEY_VAL_DELIMITER = "=";
+        if (secretMappingsStr != null && !secretMappingsStr.isEmpty()) {
+            String[] mappingList = secretMappingsStr.split(ENTRY_DELIMITER);
+            for (String eachMapping : mappingList) {
+                String[] keyValPair = eachMapping.split(KEY_VAL_DELIMITER);
+                String key = keyValPair[0].trim();
+                String value = keyValPair[1].trim();
+                if (key.isEmpty() || value.isEmpty()) {
+                    logMetacat.debug(
```
taojing2002, line 472: May we increase the debug level to error, which is the default debug level in Metacat? So operators can find the issue on the log file easily.

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
artntek: This was my question that we discussed in our backend meeting yesterday. The problem with passing these values to d1 Settings is that the current code only accepts a file path string, and reads the settings from a file on disk, which is the security issue we're trying to avoid by using env variables.

From our discussion, it sounded like none of the secrets are currently needed in the d1 code, so there is no real need to pass the vars there, yet. I'm therefore proposing waiting until we need to do this (if we ever do), before adding that complexity.

I added a note in the javadoc for this method (see line 308, above) and also in [metacat.properties on line 55](https://github.com/NCEAS/metacat/pull/1636/files/3a4005c783df382c7bf5c0326edb1fc6baa97e53#diff-3c44a421a363e2e43cea038a031a838831a483d6b78218b009bd7751b293e393R55), so other devs know the values aren't passed through to d1.  

Are you OK with this approach?

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
artntek: good idea - will do!

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
taojing2002: Yeah, I forgot that. But in the `Setting` class, there is a method `getConfiguration()` and it will return the object `configuration` which holds the property values. We may call the method `setProperty` to add the new key/value pairs. But I am not sure if the setProperty will store the key/value pair into a file. If it does, I don't where it stores. Here is the link of documentation of the [method](https://commons.apache.org/proper/commons-configuration/apidocs/org/apache/commons/configuration2/AbstractConfiguration.html#setProperty(java.lang.String,java.lang.Object)).
Your solution seems fine to me too. The above is only my thought.

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
artntek: (agreed in slack that WARN was more appropriate, since it's not truly an error condition)

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
artntek: added implementation notes as guidance for our future selves, if this becomes necessary

taojing2002 (approved):

Issue #1623, opened 2023-05-11
Metacat is deployable on Kubernetes, using a Helm Chart


artntek:
Built on top of the work for #1218 (**build a docker image for metacat**) - the goal is to be able to use helm to install or upgrade metacat on a k8s cluster. 

## Issues related to this epic:
* ~~#1635~~ 
* ~~#1662~~
* #1665

artntek:
## ToDo Items

* defining site properties behavior for new vs. existing helm deployments
* some test failures to address
* Persistent Volume mounted at `/var/metacat` - working, but not poss to reconnect to same PV if Claim deleted when deployment deleted - so now creating claim only once, outside of deployment. Need to review with @mbjones 
* using secrets for passwords
* stop webapp running as root
* use helm subcharts & feature flags for dependencies (postgres, solr?)

mbjones:
Hey @artntek nice to see your progress on the Postgres sub chart over in that branch. A small thought -- I've been thinking it would be better to not commit files, and particularly binary files, in the `charts` directory, with a similar rationale to not committing other build artifacts like jar files. Otherwise, our repo is going to start accumulating versions of the pg chart as we upgrade versions. Better to stick with just the Chart.yaml defining the dependency.

For context: https://stackoverflow.com/q/73599035

When the chart gets built with `helm dependency builds the dependency will be resolved, and when we package and release it it will be included in the artifact from `helm package`. Does that resonate with you?

artntek:
Totally! I was in two minds about this, and ended up following the example I found in dataone_indexer (which has a solr binary checked in). I’d feel better leaving it out. What about the lock file - should that be checked in?

mbjones:
Yeah, sorry about the indexer -- I've been meaning to change that one too. I'm not sure about the lockfile. It's a good question. I think it is probably not needed and will be generated anyways. Putting it in place would lock upgrades to a specific version for anyone who installed the chart, even if the chart label requested is fairly generic. If we have a really specific chart version listed in the dependencies, then the lock is probably not doing anything different.

Let's add `charts` to the .gitignore so it doesn't arise for others.

artntek:
Removed binary and lock file in https://github.com/NCEAS/metacat/pull/1651

> Let's add `charts` to the .gitignore so it doesn't arise for others.

Done - https://github.com/NCEAS/metacat/pull/1652

artntek:
# Exclusions for 3.0.0+

* Don't need `cgi-bin` setup
[...]

Issue #1218, opened 2018-03-05
build a docker image for metacat


mbjones:
Create a configuration that allows Metacat to start up as a docker image with a simple command.  An example deployment would include three containers:

- [ ] a *tomcat*-based container with Metacat and MetacatUI installed
- [ ] a *postgres* container to run the metacat database
- [ ] an apache *httpd* container configured with SSL as a proxy using Let's Encrypt

Of course, the metacat image itself could be used in other deployment contexts by providing environment variables that configure it to connect to the remote database.

mbjones:
Be sure to check thread in #1226 for some excellent suggestions.

mbjones:
Consider whether to add a Kubernetes deployment descriptor as well to make it easy to deploy there.

mbjones:
Initial start on this functionality is in the `feature-docker` branch (https://github.com/NCEAS/metacat/tree/feature-docker), but that needs to be merged with the many release changes that have happened recently.

artntek:
I've decide k8s and helm work are getting off-topic for **build a docker image for metacat #1218**, so I'll leave this issue to describe the docker work ONLY, and have created a new issue for the k8s & helm work: #1623 - **Metacat is deployable on Kubernetes, using a Helm Chart**

artntek:
### TODO items on the Docker work

* create a GitHub action for building docker images and storing them in GHCR
* Alternative method of applying schema upgrades in Docker? Or apply via helm hooks? (see [flyway](https://flywaydb.org/))
* Make metacatui install optional (and maybe do separately with helm?)

artntek:
Closing this. Remaining TODO items are already represented in #1623 and related issues

PR #1637, opened 2023-06-07
branch: `feature 1218 docker image` - makes Metacat deployable on Kubernetes, using a Helm Chart

artntek:
See Issue #1623 

Built on top of the work for https://github.com/NCEAS/metacat/issues/1218 (build a docker image for metacat) - the goal is to be able to use helm to install or upgrade metacat on a k8s cluster.


mbjones (approved):
I've been following along as this goes, and looks good to me, although I haven't yet run the code. Since the PR is all contained outside of regular metacat source, I will just approve the merge and then you can refactor later if/when we find stuff that needs to change.

PR #1640, opened 2023-06-13
Revisit Feature #1635 to clarify logging and make levels more appropriate

artntek:
Revisit Feature #1635 to clarify logging and make levels more appropriate

### DIFF
--- a/src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
+++ b/src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
@@ -143,7 +143,7 @@ public class PropertiesWrapper {
             String envKey = envSecretKeyMappings.getProperty(propertyName);
             returnVal = System.getenv(envKey);
             if (returnVal == null || returnVal.trim().isEmpty()) {
-                logMetacat.warn("Didn't find env secret: " + envKey
+                logMetacat.info("Didn't find env secret: " + envKey
                                     + "; falling back to property: " + propertyName);
             } else {
                 logMetacat.info(
--- a/src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
+++ b/src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
@@ -140,16 +140,21 @@ public class PropertiesWrapper {
 
         String returnVal = null;
         if (envSecretKeyMappings.containsKey(propertyName)) {
-            returnVal = System.getenv(envSecretKeyMappings.getProperty(propertyName));
+            String envKey = envSecretKeyMappings.getProperty(propertyName);
+            returnVal = System.getenv(envKey);
+            if (returnVal == null || returnVal.trim().isEmpty()) {
+                logMetacat.warn("Didn't find env secret: " + envKey
+                                    + "; falling back to property: " + propertyName);
+            } else {
+                logMetacat.info(
+                    "Using env var secret: " + envKey + " to override property: " + propertyName);
+            }
         }
-        // System.getenv(key) returns null if key not present, or empty string if defined but empty
-        // in either case, we want to try fallback:
         if (returnVal == null || returnVal.trim().isEmpty()) {
             returnVal = mainProperties.getProperty(propertyName);
         }
-        // getProperty() returning an empty value is valid, so only disallow null:
-        if (returnVal == null) {
-            logMetacat.info("did not find the property with key " + propertyName);
+        if (returnVal == null) {  // getProperty() returning an empty string is valid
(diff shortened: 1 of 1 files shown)


src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -140,16 +140,21 @@ protected String getProperty(String propertyName) throws PropertyNotFoundExcepti
...
         String returnVal = null;
         if (envSecretKeyMappings.containsKey(propertyName)) {
-            returnVal = System.getenv(envSecretKeyMappings.getProperty(propertyName));
+            String envKey = envSecretKeyMappings.getProperty(propertyName);
+            returnVal = System.getenv(envKey);
+            if (returnVal == null || returnVal.trim().isEmpty()) {
+                logMetacat.warn("Didn't find env secret: " + envKey
+                                    + "; falling back to property: " + propertyName);
+            } else {
+                logMetacat.info(
```
artntek, line 149: I'm open to making this "warn" instead, since it might help admins track down auth problems - might be better than just :silently" overriding the properties value? What do you think, @taojing2002 and @mbjones ?

mbjones: I think WARN is an indication of a problem, and there isn't a problem here AFAICT -- the env property is being used as intended. So, logging it as INFO means there is a way to see it in the logs if someone were debugging a problem, but its a low enough level that it won't get flagged as something that needs to be examined with every WARNING.  See my slack message on where I think the right levels are for typical message types.

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -140,16 +140,21 @@ protected String getProperty(String propertyName) throws PropertyNotFoundExcepti
...
-        // System.getenv(key) returns null if key not present, or empty string if defined but empty
-        // in either case, we want to try fallback:
         if (returnVal == null || returnVal.trim().isEmpty()) {
             returnVal = mainProperties.getProperty(propertyName);
         }
-        // getProperty() returning an empty value is valid, so only disallow null:
-        if (returnVal == null) {
-            logMetacat.info("did not find the property with key " + propertyName);
+        if (returnVal == null) {  // getProperty() returning an empty string is valid
+            logMetacat.error("did not find the property with key " + propertyName);
```
mbjones, line 157: Error seems right to me here. If there is a call for a property name that doesn't exist, that likely indicates either a code or config error.

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -481,21 +486,19 @@ private void initializeSecretsFromEnvVars() {
                 String key = keyValPair[0].trim();
                 String value = keyValPair[1].trim();
                 if (key.isEmpty() || value.isEmpty()) {
-                    logMetacat.warn(
-                        "PropertiesWrapper.initializeSecretsFromEnvVars(): badly-formed entry: ("
-                            + key + ") = (" + value + ")");
+                    logMetacat.error(
+                        "Badly-formed entry in 'application.envSecretKeys' property: ("
+                            + key + ") = (" + value + "); skipping");
```
mbjones, line 491: Same here, error seems appropriate to flag a misconfiguration.

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
```diff
@@ -481,21 +486,19 @@ private void initializeSecretsFromEnvVars() {
...
                     continue;
                 }
-                logMetacat.warn(
-                    "envSecretKeyMappings: If env var: " + value
+                logMetacat.info("envSecretKeyMappings: If env var: " + value
```
mbjones, line 494: And I like INFO here as well, for normal operations.

mbjones (approved):
Looks good to me, made a few comments inline about log levels I particularly support.

src/edu/ucsb/nceas/metacat/properties/PropertiesWrapper.java
artntek: Thanks! I also struggled with the level for: `logMetacat.warn("Didn't find env secret...`. I'd be fine with making it "info" if Jing agrees (since either way, it would still require a log-level change for it to be displayed). What do you think, @taojing2002? You get the deciding vote :-)

taojing2002: The `info` level sounds good to me. @artntek

taojing2002 (approved):

PR #1643, opened 2023-06-15
Bugfix 1638b fix false negative

artntek:
MetacatAdminTest was previously giving false "pass" status, because:

1. it didn't account for one of the properties (`configutil.upgrade.solr.status`) when testing - this should have made it always fail, but...

2. it didn't validate starting conditions, so if the operator had ever run the admin config/updates, the test would pass because `configutil.upgrade.solr.status` would always start with the value "success", thus negating (1) and allowing the test to pass

This PR fixes the above by:
1. Adding `configutil.upgrade.solr.status` to the existing tests
2. Adding validation for the status of all the `configutil.upgrade.*` properties before the tests starts
3. Adding validation for the status of all the `configutil.upgrade.*` properties at each step
4. Adding a test that will fail if anyone adds another `configutil.upgrade.*` value to the properties file without updating the test to include it.

### DIFF
--- a/test/edu/ucsb/nceas/metacat/admin/MetacatAdminTest.java
+++ b/test/edu/ucsb/nceas/metacat/admin/MetacatAdminTest.java
@@ -56,75 +56,76 @@ public class MetacatAdminTest {
         assertFalse(incompletePropsList.containsAll(names));
     }
 
-  /**
-   * Test the method of updateUpgradeStatus. Full list as of June 2023:
-   * <code>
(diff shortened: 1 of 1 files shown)


taojing2002 (approved):
Looks good to me. Thanks!

taojing2002:
I looked at the blame for the metacat.properties file and found `configutil.upgrade.solr.status` was added later than other update status. So it explains why we didn't add the test for this property. The test always passed and I didn't find we need to add more :)

PR #1645, opened 2023-06-21
Feature 1623 - helm: separating `administrator.username` & `auth.administrators` properties

artntek:
related to Issue #1623 

I originally conflated the admin username (used for DB upgrades at launch) with the `auth.administrators` property, when cleaning up the dockerfile and entrypoint script (i.e. I cleaned it up just a little too much :-) This meant that we could no longer provide a colon-delimited list of allowed admins for `auth.administrators`.

This PR undoes that, so now we have:

In `values.yaml:

* `administrator.username` and
* `auth.administrators`

as separate values, and edits to `statefulset.yaml`, `secrets.yaml` and `metacat.properties` to support these changes

Finally, I found a bug resulting from the recent change to the docker base image (`tomcat:8.5.90-jre8-temurin-jammy`) - this necessitated an upgrade to python 3, which broke the password hashing in `docker-entrypoint.sh`, so that's been fixed too.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -1,18 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-if [ "$1" = 'catalina.sh' ]; then
-
-    if [ -z "$METACAT_AUTH_ADMINISTRATORS" ] ||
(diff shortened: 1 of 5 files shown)


taojing2002 (approved):
I just looked the metacat.properties file and it is safe to make this change.

PR #1646, opened 2023-06-22
stopped aborting startup when in container/k8s if site props is r/o 

artntek:
stopped aborting startup when in container/k8s if site props is r/o (expected in that environment)

### DIFF
--- a/src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
+++ b/src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
@@ -38,12 +38,14 @@ import java.util.Set;
  * properties file is expected to be a read-only configMap in kubernetes, so the create/write
  * checks should be skipped.
  * </p><p>
- * TODO: Add more test cases - this initial implementation is minimum viable solution for
- *  /var/metacat not being writeable when metacat-site.properties is not available.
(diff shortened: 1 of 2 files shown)


taojing2002 (approved):
It looks good!

mbjones (approved):
Took a quick look, overall looks good. 

In the comments you state that this checks if `/var/metacat` is writable, but I don't immediately see where that is checked. It would be good to check it, as even in k8s we need that writable volume to be in place for data to be written.

artntek:
> In the comments you state that this checks if `/var/metacat` is writable, but I don't immediately see where that is checked. It would be good to check it, as even in k8s we need that writable volume to be in place for data to be written.

Unfortunately, the comments are misleading. When the app is starting up, we don't know for sure that /var/metacat/ is the location for the metacat files (since that's configurable). (We do know it's /var/metacat in k8s, because that's hardcoded into the image, but we also create those directories in `docker-entrypoint.sh`, thus guaranteeing they are there (or startup fails).

If you want me to add validation for /var/metacat being writeable only when running in a container, I'm happy to do that, but since that's checked elsewhere, a simpler approach would be to change the comments so they aren't misleading, and leave the code as is. 

LMK which you prefer.

mbjones:
Fixing the comments for now is fine. As you say in that class documentation, there are more checks to be added later.

PR #1654, opened 2023-06-29
Bugfix 1627 cancel button change properties

taojing2002:
In the routine to handle the first display/cancel/error scenario, it brutally set some properties (including `solr-home`) to the default values in the memory and also save to the property file. Now I added code to allow the settings only happen when the properties don't have values. This change fixes the issue.

### DIFF
--- a/src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
+++ b/src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
@@ -133,7 +133,7 @@ public class PropertiesAdmin extends MetacatAdmin {
                     PropertyService.setPropertyNoPersist(applicationTempDir,
                             externalDir + FileUtil.getFS() + "temporary");
                 }
-                String replicationLogDir = "replication.logdir";
+                final String replicationLogDir = "replication.logdir";
                 if (isNotSet(replicationLogDir)) {
                     PropertyService.setPropertyNoPersist(replicationLogDir,
                             externalDir + FileUtil.getFS() + "logs");
--- a/src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
+++ b/src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
@@ -88,116 +88,61 @@ public class PropertiesAdmin extends MetacatAdmin {
 
                 // Attempt to discover the following properties.  These will show
                 // up in the configuration fields if nothing else is provided.
-                String applicationContext = null;
-                try {
-                    applicationContext = PropertyService.getProperty("application.context");
-                } catch (PropertyNotFoundException ee) {
-                    applicationContext = null;
-                }
-                if (applicationContext == null || applicationContext.trim().equals("")) {
-                    PropertyService.setPropertyNoPersist("application.context",
+                final String applicationContext = "application.context";
+                if (isNotSet(applicationContext)) {
+                    PropertyService.setPropertyNoPersist(applicationContext,
                             ServiceService.getRealApplicationContext());
                 }
-                String serverName = null;
-                try {
-                    serverName = PropertyService.getProperty("server.name");
-                } catch (PropertyNotFoundException ee) {
-                    serverName = null;
+                final String serverName = "server.name";
+                if (isNotSet(serverName)) {
+                    PropertyService.setPropertyNoPersist(serverName, SystemUtil
+                            .discoverServerName(request));
                 }
-                if (serverName == null || serverName.trim().equals("")) {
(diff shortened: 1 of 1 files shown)


src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
```diff
@@ -88,29 +88,119 @@ public void configureProperties(HttpServletRequest request,
...
+                String applicationContext = null;
+                try {
+                    applicationContext = PropertyService.getProperty("application.context");
+                } catch (PropertyNotFoundException ee) {
+                    applicationContext = null;
+                }
+                if (applicationContext == null || applicationContext.trim().equals("")) {
+                    PropertyService.setPropertyNoPersist("application.context",
+                            ServiceService.getRealApplicationContext());
+                }
```
artntek, line 100: (applies to lines 91 thru' 200)

Instead of repeating this pattern multiple times, can we refactor it out into a utility method? That way, this code block would be much more readable and less error-prone for future editing. Here's an example:

1. First define a method, e.g.:
```
            private boolean isNotSet(String propertyKey) {
                String property = null;
                try {
                    property = PropertyService.getProperty(propertyKey);
                } catch (PropertyNotFoundException ee) {
                    property = null;
                }
                return !(property == null || property.trim().equals(""));
            }
```

2. then instead of having this for each property:
```
                String applicationContext = null;
                try {
                    applicationContext = PropertyService.getProperty("application.context");
                } catch (PropertyNotFoundException ee) {
                    applicationContext = null;
                }
                if (applicationContext == null || applicationContext.trim().equals("")) {
                    PropertyService.setPropertyNoPersist("application.context",
                            ServiceService.getRealApplicationContext());
                }
```
...we can simplify to:
```
                if (isNotSet("application.context") {
                    PropertyService.setPropertyNoPersist("application.context",
                            ServiceService.getRealApplicationContext());
                }
```

src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
```diff
@@ -251,7 +341,7 @@ public void configureProperties(HttpServletRequest request,
                     SolrVersionChecker checker = new SolrVersionChecker();
                     isOldVersion = checker.isVersion_3_4(solrHomePath);
                 } catch (Exception e) {
-                    logMetacat.warn("PropertiesAdmin.configureProperties - we can't determine if the given directory is a old version of solr  since "+e.getMessage()+". But we consider it is not an old version.");
+                    logMetacat.warn("PropertiesAdmin.configureProperties - we can't determine if the given solr home directory is a old version of solr  since "+e.getMessage()+". But we consider it is not an old version.");
```
artntek, line 344: Please wrap at 100 char column.

Suggested rewording:

we can't determine if the given solr home directory ~~is a~~ `contains an` old version of solr`,` since "` `+` `e.getMessage()` `+` `". ~~But we consider~~ `We will assume` it is not an old version.

artntek (changes requested):

src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
taojing2002: Matthew, do you think in the `isNotSet` method, the return statement should be `return (property == null || property.trim().equals(""));` rather than `return !(property == null || property.trim().equals(""));`?

artntek: you are correct - sorry about that!

artntek (approved):

src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
```diff
@@ -88,29 +88,64 @@ public void configureProperties(HttpServletRequest request,
...
+                if (isNotSet(applicationDocumentDir)) {
+                    PropertyService.setPropertyNoPersist(applicationDocumentDir,
+                            externalDir + FileUtil.getFS() + "documents");
+                }
+                final String applicationTempDir = "application.tempDir";
+                if (isNotSet(applicationTempDir)) {
+                    PropertyService.setPropertyNoPersist(applicationTempDir,
+                            externalDir + FileUtil.getFS() + "temporary");
+                }
+                String replicationLogDir = "replication.logdir";
```
artntek, line 136: I think you missed a `final` here?

artntek (approved):

src/edu/ucsb/nceas/metacat/admin/PropertiesAdmin.java
taojing2002: Yeah, i just added it. Thanks.

PR #1656, opened 2023-07-04
added solr startup check. Refactored to make props available

artntek:
* added solr startup check. 
* Also refactored to make properties available for this solr check (and for future checks). 
* Added tests

### DIFF
--- a/src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
+++ b/src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
@@ -4,6 +4,9 @@ import edu.ucsb.nceas.metacat.properties.PropertyService;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.nio.charset.MalformedInputException;
 
 import javax.servlet.ServletContextEvent;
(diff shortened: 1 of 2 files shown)


src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
```diff
@@ -85,24 +96,28 @@ public void contextDestroyed(ServletContextEvent sce) {
...
-            metacatProperties.load(Files.newBufferedReader(defaultPropsFilePath));
-            metacatProperties.store(Files.newBufferedWriter(defaultPropsFilePath), "");
-
-            sitePropsFilePath = Paths.get(
-                metacatProperties.getProperty(PropertyService.SITE_PROPERTIES_DIR_PATH_KEY),
-                PropertyService.SITE_PROPERTIES_FILENAME);
-
+            // can read?
+            defaultProperties.load(Files.newBufferedReader(defaultPropsFilePath));
+            // can write? Use isWriteable() so we don't mess up props file formatting for Jing :-)
```
taojing2002, line 112: Thanks. But the metacat.properties file is still changed  :).  I believe it happens when we write some properties (such as status) into the file.

src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
```diff
@@ -228,21 +297,23 @@ protected void abort(String message, Exception e) throws RuntimeException {
...
      * @throws RuntimeException if any unrecoverable problems are found that should cause startup
      *                          to be aborted
      */
-    private void validateSitePropertiesFileRwAccess(Path sitePropsFilePath)
-        throws RuntimeException {
+    private void validateSitePropertiesFileRwAccess(
+        Path sitePropsFilePath, Properties defaultProperties) throws RuntimeException {
 
-        Properties siteProperties = new Properties();
+        runtimeProperties = new Properties(defaultProperties);
```
taojing2002, line 311: Is there any reason why we need the `defaultProperties` to initialize the runtimeProperties? If we just use the default constructor to initialize the runtimeProperties object, we can only has one parameter in the method signature.

src/edu/ucsb/nceas/metacat/healthchecks/StartupRequirementsListener.java
artntek: agreed - I just thought I would be better not to have it happen in yet another place :-)

taojing2002 (approved):
Looks good. Thank you!

Issue #1630, opened 2023-05-27
Add new/updated documentation explaining how the properties overlay system works


artntek:
Related to epic #1611 

### As a fast follow-up [~~after the properties overlay PR is merged~~](https://github.com/NCEAS/metacat/pull/1622)

Add new/updated documentation explaining how the properties system works, and how the properties files get loaded in the standalone and the k8s deployment scenarios (see [google doc](https://docs.google.com/document/d/1nm2f8dDKJvUKGQ2OQdIz_jldk0-jlYn_jfLD-o2n-JM/edit?usp=sharing) for details).
- (don't forget to mention site props path is saved in metacat.properties, and the implications of this)

Requires, but is not limited to:
* Changes to [the admin guide](https://knb.ecoinformatics.org/knb/docs/) (in restructured text format; admin guide gets built for each release)
  * [5. Configuring Metacat](https://knb.ecoinformatics.org/knb/docs/configuration.html#configuring-metacat) - especially:
    * [5.3.1. Global Properties (server, ports, etc)](https://knb.ecoinformatics.org/knb/docs/configuration.html#global-properties-server-ports-etc) - new screenshot for site properties path field
    * [5.4. Additional Configuration](https://knb.ecoinformatics.org/knb/docs/configuration.html#additional-configuration)
    * [17. Appendix: Metacat Properties](https://knb.ecoinformatics.org/knb/docs/metacat-properties.html#appendix-metacat-properties)
  * ...but grep **everywhere** for `metacat.properties` and variations on property, properties etc
* add README in `metacat/test/`
* `Contributing.md` (test properties vs. prod properties overlay, and reference `metacat/test/README`)
* also update `Send feedback and bugs to: http://bugzilla.ecoinformatics.org/` - we no longer use bugzilla. Substitute: "Create an Issue at https://github.com/NCEAS/metacat/issues/new"

PR #1660, opened 2023-07-11
Feature 1630 documenting properties overlay

artntek:
See #1630 

The main differences are:
* Index is 3 layers deep instead of 2. Longer, but helps with discoverability?
* Updated 5.4.1. Metacat Properties Overview
* Added 5.4.4. Read-only Mode (Jing's page)
* Updated intro to 16. Appendix: Metacat Properties
* Added new page at 17.1. Metacat Configuration Properties System

(I touched a bunch of other files that were causing build warnings and had typos etc, and to update references to editing the metacat.properties file)


artntek:
@taojing2002 & @mbjones - I think we agreed that I could merge docs to develop without review, and I'd just tag you on the PRs. However, there's no rush on this, so I'll leave it a few days before merging, in case you want to take a look. Worst case, I can always fix stuff after merge, since it' sonly documentation. Thanks!

PR #1661, opened 2023-07-12
Bugfix 1630 docs sphinx

artntek:
metacat.properties: re-indenting wrapped lines to un-break python config for sphynx

metacat.properties.metadata.xml: removed references to "knb install direcories" from description labels on admin form


taojing2002 (approved):
Looks good!

Issue #1662, opened 2023-07-19
Support DataONE Member Node Configuration when running in k8s


artntek:
Part of Epic #1623
related to #1665

Requested/identified by Val@ESSDIVE
Admin UI: DataONE Configuration - allow submitting settings to CN without saving locally, so functionality is not restricted by read-only properties in k8s

artntek:
## Notes

Jing:
Metacat will compile a node document (xml) based on the form which the operator fills out and sends the node document to cn.
So the best way is to look at the code. I believe it is in MNodeService class and the method name is getCapacities.
If we don’t modify those properties locally. I am wondering if it creates inconsistence between node documents in cn and in member node itself.

artntek:
Proposed logic for the registration and update processes, after discussions with @mbjones and @taojing2002:

```mermaid
flowchart TD

k8s(["Kuberneters Startup, or\nLegacy Admin Page Submit"])
--> START{"`**Previously 
        Registered as a 
        DataONE Member 
        Node (MN)?**`"}
    START -->|YES| IDCr
    START -->|NO| IDC

subgraph PREREG ["Existing Registration"]
    IDCr{"`**nodeId changed?**
                    (*yaml:nodeId !=
                    DB nodeId_history latest?*)`"}
    IDCr -->|YES| AUTOr{"`*K8s ONLY:*
                    Registration 
                    Change Requested?
                    (*yaml flag == 
                    today's date?*)`"}
        AUTOr -->|YES| CERTr{"`New
                    NodeId 
                    Matches Client 
                    Cert?`"}
            CERTr -->|YES| UPDCNr["`UPDATE @ CN
                                    *Push MNconfig to CN.
                                    (nodeId CHANGED)*`"]
                UPDCNr --> SUCCESSr{"`Update
                                    successful?`"}
                SUCCESSr -->|YES| UPDBr["`UPDATE DB
                                        *(authoritativeMNId
                                        & nodeId history)*`"]
                SUCCESSr -->|NO| ERRORr
            CERTr -->|NO| ERRORr["`LOG ERROR`"]
        AUTOr -->|NO| ERRORr 
    IDCr -->|NO| UPCNr["`UPDATE @ CN
                        *(Push MN config to CN.
                        (nodeId UNCHANGED)*`"]
    ERRORr & UPCNr & UPDBr --> DONEr(["Done"])
end

subgraph UNREG ["NOT Previously Registered"]
    IDC{"`**nodeId changed?**
[...]

PR #1682, opened 2023-08-30
Feature 1665 membernode upreg

artntek:
Implementation of the flow diagram shown in issue #1662


src/edu/ucsb/nceas/metacat/IdentifierManager.java
```diff
@@ -2461,45 +2461,6 @@ public boolean deleteSystemMetadata(String guid)
         return success;
     }
     
-    public void updateAuthoritativeMemberNodeId(String existingMemberNodeId, String newMemberNodeId)
```
artntek, line 2464: moved to the `D1AdminCNUpdater` class, since that's the only place it is used

src/edu/ucsb/nceas/metacat/admin/MetacatAdmin.java
```diff
@@ -89,7 +89,7 @@ public static void updateUpgradeStatus(String propertyName, String status, boole
             //this sub upgrade process failed or is in progress, so the whole process will have the same status as well
             PropertyService.setPropertyNoPersist("configutil.upgrade.status", status);
         } 
-	    if(persist) {
+	    if(persist && Boolean.parseBoolean(System.getenv("METACAT_IS_RUNNING_IN_A_CONTAINER"))) {
```
taojing2002, line 92: Is there a reason that Metacat doesn't need to persist properties when it is not running in a container?

mbjones: Good Q. Also, the env variable `METACAT_IS_RUNNING_IN_A_CONTAINER` seems far too long and verbose, and will interfere with display of the environment.  Can it be shortened to `IS_CONTAINER` or something more clever?

artntek: Does that answer your question, @taojing2002 ?

artntek: And @mbjones - when i rename it, how would you feel about the new name using “k8s” instead of the more generic “container”?

artntek: Sorry, @taojing2002 - the double negative confused me 😂 - I understand now. The condition should be NOT running in a container. I will change

taojing2002 (approved):

src/edu/ucsb/nceas/metacat/admin/MetacatAdmin.java
artntek: @mbjones - I will rename the `METACAT_IS_RUNNING_IN_A_CONTAINER` env var, but will do it as a separate PR, to keep things simpler, since I'm already working on another branch that uses it too

artntek: https://github.com/NCEAS/metacat/pull/1693

PR #1691, opened 2023-09-16
Feature 1665 k8s db auto update

artntek:
1. Pulled out Member Node initialization from metacat servlet into its own class (`K8sAdminInitializer`) and did same for unit tests
2. In this new class, also added automatic database initialization when running under k8s (and added tests for this)
3. Removed hacky `curl` method of logging into admin and running DB updates from `docker-entrypoint.sh`

### DIFF
--- a/src/edu/ucsb/nceas/metacat/startup/K8sAdminInitializer.java
+++ b/src/edu/ucsb/nceas/metacat/startup/K8sAdminInitializer.java
@@ -1,0 +1,177 @@
+package edu.ucsb.nceas.metacat.startup;
+
+import edu.ucsb.nceas.metacat.MetacatVersion;
+import edu.ucsb.nceas.metacat.admin.AdminException;
+import edu.ucsb.nceas.metacat.admin.D1Admin;
+import edu.ucsb.nceas.metacat.admin.DBAdmin;
+import edu.ucsb.nceas.metacat.database.DBVersion;
+import edu.ucsb.nceas.metacat.util.SystemUtil;
+import edu.ucsb.nceas.utilities.GeneralPropertyException;
+import edu.ucsb.nceas.utilities.PropertyNotFoundException;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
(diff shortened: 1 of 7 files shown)


src/edu/ucsb/nceas/metacat/startup/K8sAdminInitializer.java
```diff
@@ -0,0 +1,134 @@
...
+        MetacatVersion metacatVersion;
+        DBVersion dbVersion;
+        String mcVerStr = "NULL";
+        String dbVerStr = "NULL";
+        try {
+            metacatVersion = SystemUtil.getMetacatVersion();
+            mcVerStr = metacatVersion.getVersionString();
+            dbVersion = DBAdmin.getInstance().getDBVersion();
+            dbVerStr = dbVersion.getVersionString();
+            if (metacatVersion.compareTo(dbVersion) == 0) {
```
taojing2002, line 90: Do we need to consider the scenario that the metacat version is less than the db version?

artntek: I adopted the same logic as that used in [the DB admin page](https://github.com/NCEAS/metacat/blob/372a79e1222b70f9001048d6ab45ae4d420b5bfb/lib/admin/metacat-configuration.jsp#L123), since the DBAdmin::upgradeDatabase() method calls [DBAdmin.getUpdateScripts()](https://github.com/NCEAS/metacat/blob/d617d1480986ff89802ebe071de4fcbe88d2e504/src/edu/ucsb/nceas/metacat/admin/DBAdmin.java#L653), which iterates through the DB versions and figures out which updates need to be applied (and so it takes care of the cases where DB < metacat, **and** DB > Metacat). 

I think that's a good approach, because the logic is in one place (DBAdmin.getUpdateScripts()) instead of three places, and if we ever decide to support downgrades, we can do that simply by changing DBAdmin, without changing the calling code.

What do you think?

taojing2002:
Sounds good! Thank you!

taojing2002 (approved):

PR #1717, opened 2023-10-30
get indexer-related passwords from env vars

artntek:
When Metacat is using `dataone_indexer` as a subchart, both the indexer and metacat need access to the same shared passwords in k8s secrets. 

Unfortunately, Metacat uses two different properties frameworks. The Java Properties framework has already been modified to work with environment variables, but the indexer-related Metacat code uses the Apache Settings framework, which has not. 

This PR therefore makes the necessary passwords available to Metacat via env variables.


artntek:
Found out how to do this via built-in functionality in apache Settings framework - no code changes necessary

Issue #1603, opened 2023-01-30
Reimplement the feature of resubmitting failed index tasks


taojing2002:
Currently the class of IndexGeneratorTimerTask.java in  metacat-index can resubmit the failed index tasks periodically.  However, the implementation directly uses the hazelcast code, which is obsoleted in this release. We need to reimplement the class in the metacat component.

taojing2002:
We found there is a database table (i`ndex_event`) and java class  `IndexEventDAO` behind hazelcast. So now we directly use the code in Metacat.

taojing2002:
Now, the feature, which resubmits the index tasks which fails to be put into RabbitMQ (e.g., the RabbitMQ is down), has been done. 

However, we still need to think about how to resubmit the index tasks, which fail in Dataone-Index.

artntek:
see https://github.com/DataONEorg/dataone-indexer/issues/23

PR #1722, opened 2023-11-01
Feature 1436 storage and indexing

taojing2002:
This merge is the fix for this ticket:

https://github.com/NCEAS/metacat/issues/1603

Mainly we re-implemented the mechanism to re-submit index tasks which fail to be put into RabbitMQ.


src/edu/ucsb/nceas/metacat/index/queue/FailedIndexResubmitTimerTask.java
```diff
@@ -0,0 +1,176 @@
...
+import edu.ucsb.nceas.metacat.index.MetacatSolrIndex;
+
+
+/**
+ * A timer to regenerate failed index tasks or index tasks in a given time frame.
+ */
+public class FailedIndexResubmitTimerTask extends TimerTask {
+
+    public static final int WAITTIME = 10000;
+    public static final int MAXWAITNUMBER = 180;
```
artntek, line 44: I don't believe these are ever used anywhere? Can they be deleted?

src/edu/ucsb/nceas/metacat/index/queue/FailedIndexResubmitTimerTask.java
```diff
@@ -0,0 +1,176 @@
...
+    private static Log log = LogFactory.getLog(FailedIndexResubmitTimerTask.class);
+    //if this task need to reindex the previously failed index task
+    private boolean needReindexFailedEvent = true;
+    private long maxAgeOfFailedIndexTask = 864000000; // 10 days
+    
+    /**
+     * Constructor
+     * @param solrIndex
+     * @param systemMetadataListener
```
artntek, line 53: delete incorrect javadoc params

metacat-common/src/main/java/edu/ucsb/nceas/metacat/common/index/event/IndexEvent.java
```diff
@@ -34,78 +33,86 @@
...
      */
-    public Event getAction() {
+    public String getAction() {
         return this.action;
     }
-    
+
     /**
      * Set the action of the event
      * @param action
```
artntek, line 64: missing description. For example, what are valid input values?

src/edu/ucsb/nceas/metacat/index/queue/FailedIndexResubmitTimerTask.java
```diff
@@ -0,0 +1,176 @@
...
+                                    IndexGeneratorTimerTask.submitIndex(id);
+                                    //Succeeded and remove it from the index event table
+                                    IndexEventDAO.getInstance().remove(event.getIdentifier());
+                                }
+                            } catch (Exception e) {
+                                log.warn("FailedIndexResubmitTimerTask.reIndexFAiledTasks - failed "
+                                        + "to submit the reindex task for the pid " + id
+                                        + " since " + e.getMessage());
+                            }
+                        } else {
```
jeanetteclark, line 133: should something else happen in this case that the task is too old? maybe marking the task in the DB as permanently failed? otherwise it will continue to get picked up and discarded by the method

src/edu/ucsb/nceas/metacat/index/queue/IndexGeneratorTimerTask.java
```diff
@@ -318,125 +240,68 @@ private void index(List<String> metacatIds) {
...
+                since = LastReindexDateManager.getInstance().getLastProcessDate();
                 index(since);
+            } catch (IOException | ParseException | InvalidRequest | InvalidToken
+                       | NotAuthorized | NotImplemented | ServiceFailure
+                       | ClassNotFoundException | InstantiationException
+                       | IllegalAccessException | XPathExpressionException
+                       | NotFound | UnsupportedType | SolrServerException
+                       | SAXException | ParserConfigurationException | OREParserException e) {
+                log.error("IndexGenerator.run - Metadata-Index couldn't generate indexes for "
+                           + "those documents which haven't been indexed : " + e.getMessage());
```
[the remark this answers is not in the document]
mbjones: Adding exceptions to logging is a bit brutal because of the stack traces. We have far too many stack traces for inconsequential stuff as it is. My opinion is that stack traces should only be added to the log at DEBUG levels, except when a true ERROR occurs that would normally be considered a bug in Metacat. Normal operations under INFO debug levels should not produce any stack traces, and any that do show up in the logs should be treated as severe failures to be tracked down and eliminated. In this particular case, I think this is INFO level notification that a particular document wasn't indexed, and not a bug in Metacat. The key is for the log to contain enough info to understand why indexing didn't complete (which probably means differentiating the various exceptions that are rolled together here into those that are transient (and so the doc can be re-queued for indexing) versus permanent (and so need to be added to a dead-letter queue to be dealt with by the repo admins). Even better, this document-level failure to index info probably should go in a more structured data store so it can be delivered to the repo admin/users.

jeanetteclark (commented):
looks good but wondering about the case I described in my comment

src/edu/ucsb/nceas/metacat/index/queue/IndexGeneratorTimerTask.java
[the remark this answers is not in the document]
artntek: Agreed. My comment was because I saw `log.error(`. If it should instead be `log.info(`, then we're fine with no stacktrace

taojing2002:
Thanks! log.info sounds good to me.

src/edu/ucsb/nceas/metacat/index/queue/FailedIndexResubmitTimerTask.java
taojing2002: Yeah. On the line 111, we check if a task is too old. This variable of `maxAgeOfFailedIndexTask` is configuration. The default value is 20 days I believe.

jeanetteclark: yes but then if the task is too old, a warning is logged and nothing else happens as far as I can tell. so the task will get picked up again and another warning...and that will just continue to happen I think

jeanetteclark (approved):

artntek (approved):
🎉

PR #1728, opened 2023-11-08
Feature 1727 sphinx7

mbjones:
Support recent versions of python3 and sphinx 7 for admindoc builds. Test with `ant admindoc`. Older versions of sphinx should also work, starting with the first version that supported the `styles` key, but I had a hard time determining which version that was from the sphinx release notes. The python changes are to change `SafeConfigParser` (a python 2.7 fixture) which was replaced with `ConfigParser` and some methods renamed/refactored. I think any python after 3.2 would work, but I only tested against 3.10 and 3.11.


mbjones:
Oh, and by the way -- while I added the mermaid diagram from @doulikecookiedough , the build for that using `sphinx.mermaid` was not working, so I commented out the extension but left the mermaid code in the RST doc. We'll need to cycle back to getting the mermaid extension to work if we want to use it.

taojing2002 (commented):
Looks great! But do you really mean to  modify those 15 images files, Matt?

artntek (approved):

mbjones:
@taojing2002 yes, I intended to add the images. when you rerun the sphinx generator it generates new image files which are different from the old ones. So, by checking these in now, when other people regenerate them it shouldn't generate any changes.

artntek:
can we merge this, @mbjones ?

mbjones:
Yes, feel free if its working for you. It worked for me, and I felt it would improve things. But its up to you and Jing if it looks like an improvement.

PR #1735, opened 2023-11-14
Second PR from feature-1481-java-17

taojing2002:
This is a simple PR:
1. Removed two obsoleted properties which related to a SVN url in the metacat-common/pom.xml file.

2. Changed the directory from tomcat8 to tomcat9 in the build.properties file.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -11,7 +11,7 @@ metacat.releaseCandidate=
 # install ant target uses this to determine where to drop 
 # the war file in an installation.  Test classes will use
 # this value to find metacat.properties
-app.deploy.dir=/opt/local/share/java/tomcat9/webapps
+app.deploy.dir=/opt/local/share/java/tomcat/webapps
 
 # CVS access to retrieve latest EML, seek and utilities 
 cvsroot=:ext:${env.USER}@cvs.ecoinformatics.org:/cvs
--- a/build.properties
+++ b/build.properties
@@ -8,13 +8,10 @@ metacat.version=3.0.0
 metacat.releaseCandidate=
 
 # Tomcat dev deployment directory
-build.tomcat.dir=/usr/share/tomcat9
-#build.tomcat.dir=/Users/leinfelder/tools/apache-tomcat-6
 # install ant target uses this to determine where to drop 
 # the war file in an installation.  Test classes will use
 # this value to find metacat.properties
 app.deploy.dir=/opt/local/share/java/tomcat9/webapps
-#app.deploy.dir=/tmp
 
 # CVS access to retrieve latest EML, seek and utilities 
 cvsroot=:ext:${env.USER}@cvs.ecoinformatics.org:/cvs
@@ -74,8 +71,3 @@ install.ecogrid=true
 # Flag to make perl code or not. Possible values are true or false
 make.perl.code=true
 
-#ecogrid.hostname=localhost
-
-# The metacat current dir's abolute value. If you set install.ecogrid=false, this variable
-#  do NOT need to be configured
-#metacat.dir=/home/daigle/workspace/metacat
--- a/build.properties
+++ b/build.properties
@@ -8,12 +8,12 @@ metacat.version=3.0.0
 metacat.releaseCandidate=
 
 # Tomcat dev deployment directory
-build.tomcat.dir=/usr/share/tomcat8
+build.tomcat.dir=/usr/share/tomcat9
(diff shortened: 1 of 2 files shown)


build.properties
```diff
@@ -8,12 +8,12 @@ metacat.version=3.0.0
 metacat.releaseCandidate=
 
 # Tomcat dev deployment directory
-build.tomcat.dir=/usr/share/tomcat8
+build.tomcat.dir=/usr/share/tomcat9
```
artntek, line 11: Can we delete this altogether?? I could be wrong, but I don't believe this property is necessary any more? 

On my dev machine, I have not changed it from the original setting (`build.tomcat.dir=/usr/share/tomcat8`), and that path does not actually exist on my laptop, yet I've been successfully building metacat and running tests for 11 months :-) 

I looked through `build.xml`, and when I follow through the steps that use `build.tomcat.dir`, it looks like it's used in order to add all the jar files form the tomcat `lib` directory into the classpath. However, now we're using maven to manage dependencies, it seems like the compiler should find everything it needs in the lib/maven/ directory?

Let me know if I missed something

build.properties
```diff
@@ -8,12 +8,12 @@ metacat.version=3.0.0
...
 #build.tomcat.dir=/Users/leinfelder/tools/apache-tomcat-6
 # install ant target uses this to determine where to drop 
 # the war file in an installation.  Test classes will use
 # this value to find metacat.properties
-app.deploy.dir=/opt/local/share/java/tomcat8/webapps
+app.deploy.dir=/opt/local/share/java/tomcat9/webapps
 #app.deploy.dir=/tmp
```
artntek, line 17: delete

build.properties
```diff
@@ -8,12 +8,12 @@ metacat.version=3.0.0
 metacat.releaseCandidate=
 
 # Tomcat dev deployment directory
-build.tomcat.dir=/usr/share/tomcat8
+build.tomcat.dir=/usr/share/tomcat9
 #build.tomcat.dir=/Users/leinfelder/tools/apache-tomcat-6
```
artntek, line 12: delete

build.properties
```diff
@@ -8,12 +8,12 @@ metacat.version=3.0.0
...
 # install ant target uses this to determine where to drop 
 # the war file in an installation.  Test classes will use
 # this value to find metacat.properties
-app.deploy.dir=/opt/local/share/java/tomcat8/webapps
+app.deploy.dir=/opt/local/share/java/tomcat9/webapps
```
artntek, line 16: instead of having to continue editing this each time we upgrade tomcat, would it be feasible to just use a default directory name of `tomcat` instead of `tomcat9`? The user can then decide either to use a symlink, or to change this property. 

My motivation: I didn't want to change this property because git would then always show it as modified (unless I add it to .gitignore, which could lead to other problems). I therefore added a symlink, which currently looks like this - not a great experience 🤣:
```console
$ la /opt/local/share/java/
total 8
drwxr-xr-x  4 root  wheel  -  128B Sep 21 20:07 ./
drwxr-xr-x  3 root  wheel  -   96B Feb  8  2023 ../
-rw-r--r--  1 root  wheel  -  220B Sep 21 20:07 README.md
lrwxr-xr-x  1 root  wheel  -   34B Sep 21 20:04 tomcat8 -> /opt/homebrew/opt/tomcat@9/libexec
```

In case you're curious, here's what I wrote in README.md (I wrote it as a reminder after my build broke when I upgraded to tomcat 9):

```console
$ cat /opt/local/share/java/README.md

# SHAMELESS HACK!

tomcat8 points to tomcat9 because tomcat8 is used in metacat's ant build.properties:

`app.deploy.dir=/opt/local/share/java/tomcat8/webapps`

...and I don't want to keep modifying it
```

artntek (changes requested):

build.properties
taojing2002: There is no easy ways to accommodate different settings. The purpose of the build.properties files is to give users a way to customize their settings. So changing on this file is every common, just like the values.yam on k8s.

artntek: I understand that. If it is common to edit this file, that also means we can choose any default value we want.

I therefore think it makes sense to choose a default value that is not tied to a particular version of tomcat. This will reduce maintenance overhead and potential mistakes.

taojing2002: That sounds good. I will remove the version there.

artntek (approved):

PR #1742, opened 2023-11-17
Bug 1736 not found harvester registration

taojing2002:
Removed the servlets which related to the harvester.


lib/web.xml
```diff
@@ -72,56 +72,12 @@
...
-      </init-param>
-      <init-param>
-        <param-name>listings</param-name>
-        <param-value>true</param-value>
-      </init-param>
-      <load-on-startup>3</load-on-startup>
-    </servlet>
-    -->
-    
     <!--
```
artntek, line 81: Please add a very brief note saying why this is here and commented out (e.g. "This is an example for users who want to enable the OAI-PMH service - see: " ((add a link to docs))

lib/web.xml
```diff
@@ -217,104 +173,94 @@
...
+        <servlet-name>MNRestServletV2</servlet-name>
+        <url-pattern>/d1/mn/</url-pattern>
+    </servlet-mapping> 
+
+     <servlet-mapping>
+        <servlet-name>MNRestServletV2</servlet-name>
+        <url-pattern>/d1/mn</url-pattern>
+    </servlet-mapping> 
+
+    <!--
```
artntek, line 227: pls add brief note - see line 81 comment above

artntek (changes requested):
looks good - just a couple of documentation suggestions

artntek (approved):

Issue #1741, opened 2023-11-17
Clean up lib/style/skins directory


taojing2002:
Since we have some of skins which depend the path query, we need remove them.

taojing2002:
- [ ] account
- [x] dataone
- [x] default
- [x] dev
- [x] esa
- [x] first
- [x] first-assessment
- [x] first-item
- [x] kepler
- [x] knb
- [x] knb2
- [x] lter
- [x] ltss
- [ ] metacatui 
- [x] mets
- [x] nceas
- [x] nrs
- [x] obfs
- [x] parc
- [x] pisco 
- [x] saeon
- [x] sanparks
- [x] sbclter
- [x] specnet
- [x] sms
- [x] tpc-result  
- [x] tpc-run-result

The above is the list of the skins in Metacat. I think we can remove all of them except `account` and `metacatui`. The `account` skin is used for the ldap account management and `metacatui` is used for MetacatUI. The others are using the old metacat api so they can be removed. The `tpc-result` and `tpc-run-result` skins just have two simple xml files and they shouldn't work at all.

taojing2002:
I added an index.html in each skin directory to indicate that Metacat doesn't support it anymore.

PR #1744, opened 2023-11-29
Remove the advance search servlet.

taojing2002:
This is a simple PR since I just removed the advance search servlet, which used the obsoleted path query.


artntek (changes requested):
Looks good. 2 things:

1) can [this ant target](https://github.com/NCEAS/metacat/blob/4d83cfa0f5f28ee700e062b262da3f441d582246/build.xml#L616) be removed?

2) there are a lot of usages in `jsp` files within `lib/style/skins`. Can these be removed? Or better still, is it time to get rid of `lib/style/skins` altogether??

taojing2002:
1. Good catch! I removed the target on the build.xml.
2. I already input a ticket to clear out that directory two weeks ago:
https://github.com/NCEAS/metacat/issues/1741

artntek (approved):

PR #1760, opened 2023-12-07
Feature 1721 abort init fail

taojing2002:


test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
```diff
@@ -0,0 +1,152 @@
...
+ * The IT test class for the MetacatInitializer class
+ * @author tao
+ *
+ */
+public class MetacatInitializerIT {
+    private ServletContextEvent event = null;
+
+    @Before
+    public void setUp() throws Exception {
+        LeanTestUtils.initializePropertyService(LeanTestUtils.PropertiesMode.LIVE_TEST);
```
artntek, line 34: AFAICT, there's nothing in this test that requires metacat to be running. It therefore should be using PropertiesMode.UNIT_TEST. You're already mocking the `ServletContextEvent` - which is great. It looks like the only reason you are using the LIVE mode is to change the "configured" Properties for different tests. This can all be done with UNIT test mode.

(I can help figure out how to do this if you want)

test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
```diff
@@ -0,0 +1,152 @@
...
+import edu.ucsb.nceas.metacat.properties.PropertyService;
+import edu.ucsb.nceas.metacat.util.SystemUtil;
+import edu.ucsb.nceas.utilities.PropertyNotFoundException;
+
+/**
+ * The IT test class for the MetacatInitializer class
+ * @author tao
+ *
+ */
+public class MetacatInitializerIT {
```
artntek, line 29: I don't think it needs to be an IT - see below

test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Test the contextInitialized method
+     * @throws IOException
+     */
+    @Test
+    public void testContextInitialized() throws Exception {
+        //This should be a full-configured Metacat
+        MetacatInitializer metacatInitializer = new MetacatInitializer();
+        metacatInitializer.contextInitialized(event);
+        assertTrue(metacatInitializer.isFullyInitialized() == true);
```
artntek, line 51: don't need the `== true`, i.e.: `assertTrue(metacatInitializer.isFullyInitialized());`

test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
```diff
@@ -0,0 +1,152 @@
...
+        // set Metacat non-configured temporarily.
+        PropertyService
+             .setPropertyNoPersist("configutil.authConfigured", PropertyService.UNCONFIGURED);
+        //reset the fully-initialized flag to be false by creating a new instance
+        metacatInitializer = new MetacatInitializer();
+        metacatInitializer.contextInitialized(event);
+        assertTrue(metacatInitializer.isFullyInitialized() == false);
```
artntek, line 58: Can use `assertFalse`, i.e.:  `assertFalse(metacatInitializer.isFullyInitialized());

test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
```diff
@@ -0,0 +1,152 @@
...
+            assertTrue(e.getMessage().contains(Long.toString(time)));
+        }
+
+        // set Metacat non-configured temporarily.
+        // Non-configured Metacat wouldn't throw an exception.
+        PropertyService
+             .setPropertyNoPersist("configutil.databaseConfigured", PropertyService.UNCONFIGURED);
+        metacatInitializer = new MetacatInitializer();
+        metacatInitializer.contextInitialized(event);
+        assertTrue(metacatInitializer.isFullyInitialized() == false);
```
artntek, line 88: `assertFalse()`

test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
```diff
@@ -0,0 +1,152 @@
...
+            assertTrue(e.getMessage().contains(Long.toString(time)));
+        }
+
+        // set Metacat non-configured temporarily.
+        // Non-configured Metacat wouldn't throw an exception.
+        PropertyService
+             .setPropertyNoPersist("configutil.skinsConfigured", PropertyService.UNCONFIGURED);
+        metacatInitializer = new MetacatInitializer();
+        metacatInitializer.contextInitialized(event);
+        assertTrue(metacatInitializer.isFullyInitialized() == false);
```
artntek, line 122: `assertFalse()`

artntek (changes requested):

test/edu/ucsb/nceas/metacat/startup/MetacatInitializerIT.java
taojing2002: This test needs Postgresql is running and configure, needs rabbitmq is running and et al. So it wouldn't work of the box. So I call it `IT`.

src/edu/ucsb/nceas/metacat/startup/MetacatInitializer.java
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * A class to initialize Metacat
+ * @author tao
+ *
+ */
+@WebListener
+public class MetacatInitializer implements ServletContextListener{
+    public static final String APPLICATION_NAME = "metacat";
+    private static boolean fullInit = false;
+    private static Log logMetacat = LogFactory.getLog(MetacatInitializer.class);
```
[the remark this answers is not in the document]
taojing2002: It can be final. But do you think it is necessary?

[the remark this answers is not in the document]
artntek: no - you're right. I had always incorrectly assumed that declaring things final helped the compiled do some optimizations that helped with performance. However, when you asked this question, I did a bit of googling, and it appears that the performance benefits are not that great - e.g. see https://www.baeldung.com/java-final-performance#:~:text=Besides%20performance%2C%20we%20might%20consider,a%20route%20to%20immutable%20objects

So please ignore this comment! :-)

artntek (approved):

PR #1764, opened 2023-12-21
script to list and back up all sysmeta that doesn't have an identifie…

artntek:
scripts to back up and clean up orphaned sysmeta due to Chrome 120 bug - one to list & back up, and another to delete

### DIFF
--- a/src/scripts/bash/Chrome120_sysmetaclean_1_CREATE_BACKUPS.sh
+++ b/src/scripts/bash/Chrome120_sysmetaclean_1_CREATE_BACKUPS.sh
@@ -14,8 +14,8 @@ if [ -z "$TOKEN" ]; then
     exit 1
 fi
 if [ -z "$PGPASSWORD" ]; then
-    echo "Requires 'PGPASSWORD' env variable containing password for metacat db user"
-    exit 2
+    echo "WARNING: Optional 'PGPASSWORD' env variable containing database user password not set"
+    echo "You may get prompted to enter the password"
 fi
 if [ -z "$METACAT_BASE_URL" ]; then
     echo "Requires 'METACAT_BASE_URL' env variable containing the protocol, host, optional"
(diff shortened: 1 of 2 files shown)


jeanetteclark:
Hey @artntek - I did some testing of the delete script and I think we are going to have to sort out a potential edge case before running it. Since some objects were updated (and not created), some of the pids that are returned in the query to have the sysmeta deleted exist as `obsoletedBy` in the sysmeta of another pid. So we need to also modify the sysmeta of the previous pid in the version chain to set it to NA.

My procedure for investigating this was (in R since it's faster for me)

1. Take the list of pids returned from the query and get `obsoletes` fields for all of them
2. Check to see if any of those `obsoletes` are not NA
3. For those pid(s), check the sysmeta of the `obsoletes` pid to see if `obsoletedBy` is not NA
4. Sadly confirm that `obsoletedBy` is the mangled pid

If you want to investigate the pid on Arctic, the head of the version chain is `urn:uuid:2adb9ac2-2bd7-4f9a-ad24-a1ff81890786`. The "object" it obsoletes, `urn:uuid:d4ef8182-d245-4ef3-966d-6c8cfe03b5fe` is also mangled.

I'm actually not sure of the best way to repair this. I could patch together a version chain and we can pretend that intermediate pids without an actual objects associated with it aren't a problem (probably fine from a user perspective but not best practice maybe). We could delete the sysmeta for the top two pids of the chain, and modify the sysmeta for the 3rd pid.

One last wrinkle would be to see what of this synced to the CN, and making sure we also make the same modifications there.

@mbjones do you think I missed anything? As much as I want to get this out to ESS-DIVE as well, particularly with CN stuff involved we should get Jing in on the conversation. Maybe we have ESS-DIVE run the query script only and check for the obsoletes issue above ourselves to see what they might be dealing with

jeanetteclark (changes requested):
see my really long comment which probably should have gone here

jeanetteclark:
I changed the delete script such that only pids that have NULL obsoletes and obsoletedBy fields in the sysmeta are deleted

mbjones:
I went through our repositories, and summarized their state:

- ADC: 74 stranded sysmetas, 2 part of version chains (Jeanette is fixing these up)
- KNB: 30 stranded sysmetas, none part of version chains, but three older resource maps have stranded sysmetas back from July to be checked
- SCTLD: 12 stranded sysmetas, none part of version chains
- OPC: 6 stranded sysmetas, none part of version chains
- SFWMD: 4 stranded sysmetas, none part of version chains
- CERP: 0 stranded sysmetas
- DRP: 0 stranded sysmetas

jeanetteclark (approved):

PR #1767, opened 2024-01-05
Feature 1741 remove skins

taojing2002:
Remove those obsoleted skins. We only keep `metacatui` and `account`. It has 600 files changed. However, most of them were deleted. There are about only two files has the really changes.


artntek (commented):
What a great start to 2024 - seeing 600 files deleted 🙌 🎉 🍾 !!!

I think there's still more code that is specific to the skins system, that we should really get rid of, if we can. (If we don't do it as part of this PR, I'm guessing it may be another 15 years ;-) Let's discuss. 

Here are some thoughts:

1. Do we need to keep all the directory structure and the index files with redirects back to `../../common/index.html`? Is there any way we can avoid that and delete all those directories?
2. Can we now remove the skins config page from the admin UI?
3. It seems like the `skin.names` property in `metacat.properties` is now redundant. Can we get rid of it altogether?
4. Can we get rid of `src/edu/ucsb/nceas/metacat/util/SkinUtil.java`?
5. Don't forget to clean up the docs (or add some specifics to the Issue) - I see references to skins in `docs/user/metacat/source/themes.rst`, for example.

artntek (approved):

PR #1794, opened 2024-02-06
Feature 1766 update id metadata

taojing2002:


src/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandler.java
```diff
@@ -1933,5 +1945,63 @@ protected void reindex(String pid) throws InvalidRequest, ServiceFailure,
...
+     * @throws ServiceFailure
+     * @throws NotAuthorized
+     * @throws InvalidRequest
+     * @throws NotImplemented
+     * @throws IOException
+     */
+    protected void updateIdMetadata(String pid) throws ServiceFailure, NotAuthorized,
+                                                      InvalidRequest, NotImplemented, IOException {
+        boolean all = false;
+        Boolean success = Boolean.TRUE;
```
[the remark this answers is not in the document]
artntek: @also applies to the `reindex` method, line 1892 onwards

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+    /**HTTP Verb GET*/
+    protected static final byte GET = 1;
+    /**HTTP Verb POST*/
+    protected static final byte POST = 2;
+    /**HTTP Verb PUT*/
+    protected static final byte PUT = 3;
+    /**HTTP Verb DELETE*/
+    protected static final byte DELETE = 4;
+    /**HTTP Verb HEAD*/
+    protected static final byte HEAD = 5;
```
artntek, line 41: Field 'HEAD' is never used

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+    private static final String EML201_NAMESPACE = "eml://ecoinformatics.org/eml-2.0.1";
+
+    /**
+     * Setup
+     */
+    @Before
+    public void setUp() throws Exception {
+        LeanTestUtils.initializePropertyService(LeanTestUtils.PropertiesMode.UNIT_TEST);
+        context = new MockServletContext(null, PATH);
+        //@TODO The filter seems not working
```
artntek, line 65: unresolved TODO?

artntek (changes requested):

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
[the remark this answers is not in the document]
taojing2002: I think the test class shouldn't share those constants. If developers change those constants in the source code class improperly,  our test class wouldn't work. But if both test class and source class share the same constants, those changes may not be caught. What do you think?

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
taojing2002: I haven't figured it out. So far the default filter works for our testing. I can image it will have some issues for the sql query test. So I defer it. Do you have any idea?

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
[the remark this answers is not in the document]
artntek: that's a good point - I agree

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
artntek: I would suggest making that more clear. For example, if it doesn't work, can you just delete the `context.addFilter("d1Filter"...` line for now? 

And maybe update this comment to clarify what you said above? e.g. instead of a TODO, maybe it could be: 

```java
// The default filter works for our testing. 
// Note, however, that the following filter does NOT seem to work:
//  context.addFilter("d1Filter", "edu.ucsb.nceas.metacat.restservice.D1URLFilter");
// This is OK for now, but may become a problem for the sql query test" 
```

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
```diff
@@ -40,9 +41,12 @@ public class MNResourceHandlerTest {
     /**HTTP Verb HEAD*/
     protected static final byte HEAD = 5;
```
artntek, line 42: HEAD is never used

artntek: Also - can't these all be `private` instead of `protected`?

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
artntek, line 1: still unresolved from last time: https://github.com/NCEAS/metacat/pull/1794#discussion_r1482136824

artntek (changes requested):

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
taojing2002: Weird. I added a comment before. This one can be used in the future. I think we have the code to use `HEAD`.

test/edu/ucsb/nceas/metacat/restservice/v2/MNResourceHandlerTest.java
artntek: (we're having some challenges with comments not showing up, so I'm pasting Jing's reply here, to keep the conversation in one place)

>
> HEAD is never used
>

(Jing said) Weird. I added a comment before. This one can be used in the future. I think we have the code to use HEAD.

artntek: In that case, I think we should add it in future, if and when it is needed, instead of keeping it here now. 

Unused code is (a) a maintenance burden, and (b) confusing to someone else who is trying to understand the tests.

taojing2002: I generally agree to your statement regarding unused code. However, this constant will be used for sure in the future. And it is a constant, not adding complexity to the code logic. When we need it in the future, we don't dig around to figure out. So we can save lots of time. What do you think?

Those constants will be used in another test classes, such as CNResourceHandlerTest. So I declare them `protected`.

artntek: TL;DR - that's fine - leave it in. My opinion has not changed, but I don't think it's worth arguing about any longer 😁

artntek (approved):

PR #1813, opened 2024-02-26
Feature 1812 config d1 portal 233

artntek:
see #1812

PR #1815, opened 2024-02-28
Feature-1804: Admin Token Path (Highlight New Properties)

doulikecookiedough:
Hi @taojing2002, can you please help review my PR for `Feature-1804`? As discussed, we have now moved the `new` indicator/badge to the `helpDescription` field in the `metacat.properties.metadata.xml` as it is a bit involved (and not necessary) to revise other classes to dynamically add this badge to a config item.

Moving forward, to add a `new` indicator - we will manually add description text that follows this format:
- `&#91;This is an example sentence.&#93; &lt;span class="new-badge"&gt;New&lt;/span&gt;`

In `properties-configuration.jsp`, I have amended the JSP code to not include left-right square brackets when it finds `new-badge` in the description.

**Example/Screenshot:**
image

### DIFF
--- a/lib/admin/admin.css
+++ b/lib/admin/admin.css
@@ -238,6 +238,7 @@ input[type="text"] {
     margin-left: 10px;
     margin-top: 0.5em;
     display: inline-block;
+    float: left;
 }
 
 .icon-plus-sign {
@@ -274,6 +275,16 @@ input[type="text"] {
     float: left;
     margin-left: 1em;
     margin-bottom: 10px;
+}
+
+.new-badge {
+    background: green;
+    color: white;
+    margin-top: 3px;
+    border-radius: 10px;
+    padding: 2px 4px;
+    display: inline-block;
+    margin-left: 0 !important;
 }
 
 
--- a/lib/admin/admin.css
+++ b/lib/admin/admin.css
@@ -280,9 +280,8 @@ input[type="text"] {
 .new-badge {
     background: green;
     color: white;
-    margin-top: 3px;
     border-radius: 10px;
-    padding: 2px 4px;
+    padding: 0px 4px;
     display: inline-block;
     margin-left: 0 !important;
 }
--- a/lib/admin/properties-configuration.jsp
+++ b/lib/admin/properties-configuration.jsp
@@ -56,7 +56,6 @@
 			Map<Integer, MetaDataGroup> groupMap = metadata.getGroups();
 			Set<Integer> groupIdSet = groupMap.keySet();
 
-
 			for (Integer groupId : groupIdSet) {
 				// for this group, display the header (group name)
 				MetaDataGroup metaDataGroup = (MetaDataGroup)groupMap.get(groupId);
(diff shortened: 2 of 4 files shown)


lib/admin/properties-configuration.jsp
taojing2002, line 1: Do we really need to make changes on this file? The changes are a mix of format changes and code changes. Would you please specify what real changes are?

lib/admin/admin.css
taojing2002, line 1: The first time when I loaded the properties configure page, the `new` icons are colorless. After I restarted tomcat, it changed to green. I am not sure if this is an issue.
Screen Shot 2024-02-28 at 6 08 49 PM

doulikecookiedough: @taojing2002 I believe this occurred because the `admin.css` file is cached on the browser for efficient loading. This happened once for me too - but I did a hard refresh to resolve it (rather than restarting Tomcat). I am unable to reproduce this issue.

lib/admin/properties-configuration.jsp
doulikecookiedough: No - we don't need to make changes here. The .JSP page adds `[` to the beginning of the `helpDescription` field and `]` to the end of it.

Other than the formatting changes, I revised the div section to not add the left and right square brackets if it finds `new-badge` in the description, and left adding these to the description itself with the html encoded character for it.
- After changes if we want to add a new indicator:
    - ```&#91;This is an example sentence.&#93; &lt;span class="new-badge"&gt;New&lt;/span&gt;```

The `new` indicator looks better to me outside of `]` but I wasn't able to find a more elegant way to apply this change. Please let me know if you would like me to revert these changes.

taojing2002:
Okay. If you do want to change the file, please replace all tabs by 4
spaces.

taojing2002:
I see. Please close the the ticket.

lib/metacat.properties.metadata.xml
```diff
@@ -395,13 +395,25 @@
...
+    <config>
+        <key>dataone.nodeToken.file</key>
+        <label>Admin Token Path</label>
+        <group>4</group>
+        <index>14</index>
+        <description>
+            &#91;The path to the admin jwt token that will be used in dataone-indexer to access the
+            private objects' system metadata.&#93; &lt;span class="new-badge"&gt;New&lt;/span&gt;
+        </description>
+        <helpFile>docs/metacat-properties.html/#admin-token</helpFile>
```
artntek, line 414: please add a description to `docs/user/metacat/source/metacat-properties.rst` so this link works - see [my recent example here](https://github.com/NCEAS/metacat/pull/1813/files)

artntek (changes requested):

lib/metacat.properties.metadata.xml
doulikecookiedough: Apologies for missing this @artntek - I have updated the `docs/user/metacat/source/metacat-properties.rst` with a new section for `dataone.nodeToken.file` along with the following description for it below - may I ask for your feedback regarding the text?
- Quick note: There is a missing period  `.` to end the last sentence, and have added this locally already. I will wait for your further feedback (to potentially include) before pushing this typo fix.

```The path to a file that contains a (JWT) token. This token will be used in the dataone-indexer to authorize access to private objects' system metadata. A valid token consists of 3 parts delimited with a period. The first section contains the header (and info about how the JWT was signed), the second  includes information on the subject (ex. id, issuer, expiry date, etc.), and the last section is the signature. ```

artntek: Thanks, @doulikecookiedough! I would suggest making the description less technical, because admins/operators do not need to know about the internals of the token, since we will issue it for them. (correct me if I'm wrong, @taojing2002)

How about:

> The path to a file that contains an authentication token. This token will be used by the dataone-indexer, to enable indexing of private datasets.

doulikecookiedough: Sounds good to me. I've pushed the update and am going to close this PR.

Issue #1828, opened 2024-03-17
Don't use the hard code https in the property of index.d1node.baseURL


taojing2002:
Now we have hard code in this  property:
```
index.d1node.baseURL=https://${server.name}/${application.context}/d1/mn
```
So if the server doesn't use http, the index will break.

doulikecookiedough:
For reference, after updating `index.d1node.baseURL`'s default value from `https` to `http` (ex. `http://${server.name}/${application.context}/d1/mn`) - I still experienced the connection refused error:

```metacat-index 20240317-09:08:15: [ERROR]: IndexWorker.indexOjbect - cannot index the task for identifier  urn:uuid:78c0a82c-e685-44a4-8909-df42c21c24b7 since class org.dataone.client.exception.ClientSideException: /Connect to localhost:80 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1] failed: Connection refused [org.dataone.cn.indexer.IndexWorker:indexOjbect:460]```

However, after hardcoding that property to `http://localhost:8080/metacat/d1/mn` appears to have resolved the issue of my test dataset not indexing. I can now view it in my local portal.

Note, my server port is set to 8080.

mbjones:
I don't understand the problem here. This is in a properties file, right? Which is by definition **not** hardcoding -- it is where we set options that can be changed without recompiling the code. Can't the deployment just change their properties file?

Secondarily, did you mean that if the server doesn't use `https`, then the index will break? I think `https` is the correct default, and it should be discouraged/difficult for people to use plain http on an externally accessible service. The only place where plain http might make sense is within a private network (such as a Kubernetes cluster).

artntek:
I think the root problem is that the property is set automatically, with a value that is incorrect when someone is running metacat without https enabled. This could be a developer, or worse, could be someone evaluating metacat. In either case, the value is not configurable via admin UI, so it's confusing when indexing fails and the installation doesn't work, causing significant time to be lost debugging and figuring out what went wrong.

The fix is super simple, and took less time than replying to this thread :-)  (PR #1829)

taojing2002:
This property was a temporary solution when dataone-indexer can't only access the system metadata through the file system. Now 3.0.0 wouldn't use hash-store, so we need to polish it more.

artntek:
As I'm looking at this now, I have a related question...

[...]

PR #1829, opened 2024-03-18
set index.d1node.baseURL to match dataone.mn.baseURL

artntek:
#1828


taojing2002 (commented):
Do we really need this property? I think it can be replaced by `dataone.mn.baseURL` after I thought more. But this need some code change in dataone-index. Since the correctness of this property is important, it will be good to check the connection of the url. If the connection of the internal server fails ( I know someone likes to disable 8080 in Tomcat), we will try the external url. If both fail, we will ask user to check the settings. What do you think?

taojing2002 (approved):

artntek:
agreed - see discussion in Issue #1828

Issue #1859, opened 2024-04-02
OstiDOIServiceIT failed


taojing2002:
I ran this test class and it failed.
```
Testcase: testPublishIdentifierPrivatePackageToPublic took 85.143 sec
  Caused an ERROR                                                                                                           
Illegal URI: https//:foo.dataone.org/v1/resolve/testPublishPrivatePackageToPublic-resourcemap.1712034621436#aggregation; Aggregation requires a protocol-based URI
org.dspace.foresite.OREException: Illegal URI: https//:foo.dataone.org/v1/resolve/testPublishPrivatePackageToPublic-resourcemap.1712034621436#aggregation; Aggregation requires a protocol-based URI
  at org.dspace.foresite.jena.AggregationJena.initialise(AggregationJena.java:117)
  at org.dspace.foresite.OREFactory.createAggregation(OREFactory.java:78)
  at org.dataone.ore.ResourceMapFactory.createResourceMap(ResourceMapFactory.java:168)
  at org.dataone.ore.ResourceMapFactory.createResourceMap(ResourceMapFactory.java:206)
  at edu.ucsb.nceas.metacat.doi.osti.OstiDOIServiceIT.testPublishIdentifierPrivatePackageToPublic(OstiDOIServiceIT.java:559)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
  at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```
```
Testcase: testPublishIdentifierPrivatePackageToPartialPublic took 85.284 sec
  Caused an ERROR
Illegal URI: https//:foo.dataone.org/v1/resolve/testPublishIdentifierPrivatePackageToPartialPublic-resourcemap.1712034706790#aggregation; Aggregation requires a protocol-based URI
org.dspace.foresite.OREException: Illegal URI: https//:foo.dataone.org/v1/resolve/testPublishIdentifierPrivatePackageToPartialPublic-resourcemap.1712034706790#aggregation; Aggregation requires a protocol-based URI
  at org.dspace.foresite.jena.AggregationJena.initialise(AggregationJena.java:117)
  at org.dspace.foresite.OREFactory.createAggregation(OREFactory.java:78)
[...]

PR #1863, opened 2024-04-02
Made Osti tests work

taojing2002:
Ref: https://github.com/NCEAS/metacat/issues/1859


test/edu/ucsb/nceas/metacat/doi/osti/OstiDOIServiceIT.java
```diff
@@ -49,52 +32,32 @@
...
-    
+    private D1NodeServiceTest d1NodeTest;
+    private MockHttpServletRequest request;
+
     /**
      * Set up the test fixtures
      * 
      * @throws Exception
      */
     @Before
```
artntek, line 56: None of these tests run "out of the box", unless you change the doi properties. Would you mind adding the following mocks, to fix this?

```java
    @Before
    public void setUp() throws Exception {
        Properties withProperties = new Properties();
        withProperties.setProperty("server.name", "UpdateDOITestMock.edu");
        withProperties.setProperty("guid.doi.enabled", "true");
        withProperties.setProperty("guid.doi.password", "apitest");
        withProperties.setProperty("guid.doi.username", "apitest");
        closeableMock = LeanTestUtils.initializeMockPropertyService(withProperties);
    }

    @After
    public void tearDown() throws Exception {
        closeableMock.close();
    }
```

artntek (changes requested):

test/edu/ucsb/nceas/metacat/doi/osti/OstiDOIServiceIT.java
taojing2002: Hi, Matthew. This test is for the OSTI server, which is used by ESS-DIVE folks. Their password is a secret so we can't put it into the code. Even though, putting ezid one in the code is not a good idea.
This is the one of three known failure test classes. It needs to contact the OSTI server for working.

artntek: Ok - instead, then, can we check on startup and `fail` with a good explanation message (like yours, ab), if doi enabled is `true` or the doi password has no value?

artntek: suggested code for checks:

```java
    @Before
    public void setUp() throws Exception {
        final String passwdMsg =
            """
            \n* * * * * * * * * * * * * * * * * * *
            DOI CREDENTIALS NOT SET!
            Test requires OSTI-specific values for
            'guid.doi.username' & 'guid.doi.password'
            in your metacat-site.properties file!
            * * * * * * * * * * * * * * * * * * *
            """;
        try {
            assertFalse(passwdMsg, PropertyService.getProperty("guid.doi.password").isBlank());
            assertNotEquals("apitest", PropertyService.getProperty("guid.doi.username"));
        } catch (PropertyNotFoundException e) {
            fail(passwdMsg);
        }
        Properties withProperties = new Properties();
        withProperties.setProperty("guid.doi.enabled", "true");
        closeableMock = LeanTestUtils.initializeMockPropertyService(withProperties);
    }

    @After
    public void tearDown() throws Exception {
        closeableMock.close();
    }
```

test/edu/ucsb/nceas/metacat/doi/osti/OstiDOIServiceIT.java
```diff
@@ -55,25 +63,47 @@ public class OstiDOIServiceIT {
...
+                """
+                \n* * * * * * * * * * * * * * * * * * *
+                DOI CREDENTIALS NOT SET!
+                Test requires OSTI-specific values for
+                'guid.doi.username' & 'guid.doi.password'
+                in your test/test.properties file!
+                * * * * * * * * * * * * * * * * * * *
+                """;
+        Properties testProperties = LeanTestUtils.getExpectedProperties();
+        String ostiName = testProperties.getProperty("guid.doi.username");
```
artntek, line 77: Better to catch the PropertyNotFound exception as shown in my suggested snippet, otherwise the setUp method will throw an exception without any explanation, if the property is not in the test props file

test/edu/ucsb/nceas/metacat/doi/osti/OstiDOIServiceIT.java
```diff
@@ -55,25 +63,47 @@ public class OstiDOIServiceIT {
...
+        String ostiPass = testProperties.getProperty("guid.doi.password");
```
artntek, line 78: same comment

artntek (changes requested):

test/edu/ucsb/nceas/metacat/doi/osti/OstiDOIServiceIT.java
taojing2002: This statement is calling a java object `Properties`. So it will not throw a PropertyNotFoundException :( . According to the java doc, it will return null if the property doesn't exist. So I check if `ostiName` is null with the error message.

test/edu/ucsb/nceas/metacat/doi/osti/OstiDOIServiceIT.java
taojing2002: Same response.

artntek (approved):

PR #1865, opened 2024-04-03
Feature 1260 doi tests prereqs

artntek:


test/test.properties
```diff
@@ -8,7 +8,6 @@ application.deployDir=/opt/local/share/java/tomcat/webapps
 test.printdebug=true
 guid.doi.enabled=true
 metacat.contextDir=/opt/local/share/java/tomcat/webapps/metacat
-guid.doi.password=apitest
```
taojing2002, line 11: May you still keep this property and keep its value blank? So it is easy for users to add it.

artntek: 09bd80fbea220991a3a1c65ecb0f6bf4da0a6802

taojing2002 (approved):

PR #1940, opened 2024-07-03
Feature 1347 hash store

taojing2002:
It is a draft PR.


src/edu/ucsb/nceas/metacat/systemmetadata/SystemMetadataManager.java
```diff
@@ -191,24 +211,15 @@ public void store(SystemMetadata sysmeta, boolean changeModifyTime, DBConnection
         if (sysmeta != null) {
             Identifier pid = sysmeta.getIdentifier();
             if (pid != null && pid.getValue() != null && !pid.getValue().trim().equals("")) {
+                if(checkLock && !lockedIds.contains(pid.getValue())) {
+                    throw new ServiceFailure(
+                        "0000",
+                        "The pid " + pid.getValue() + " is not locked before " + "storing its "
+                            + "systemmetadata. There is a bug in the code. The "
+                            + "SystemMetadataManager"
+                            + ".lock" + " method must be called before call the store method");
```
artntek, line 220: wrapping resulted in 2 unnecessary string concats

src/edu/ucsb/nceas/metacat/systemmetadata/SystemMetadataManager.java
```diff
@@ -226,53 +237,73 @@ public void store(SystemMetadata sysmeta, boolean changeModifyTime, DBConnection
...
+                            + "the system metadata for pid " + pid.getValue()
+                            + " since " + e.getMessage());
+                } catch (InvocationTargetException e) {
+                    throw new ServiceFailure("0000", "SystemMetadataManager.store - can't store "
+                        + "the system metadata for pid " + pid.getValue()
+                        + " since " + e.getMessage());
+                } catch (IllegalAccessException e) {
+                    throw new ServiceFailure("0000", "SystemMetadataManager.store - can't store "
+                        + "the system metadata for pid " + pid.getValue()
+                        + " since " + e.getMessage());
```
artntek, line 305: since all these perform almost exactly the same action, can we condense them all into one single `catch`, like:

```java
 } catch (McdbDocNotFoundException | SQLException | InvalidSystemMetadata // [...etc]
```
```

File: src/edu/ucsb/nceas/metacat/systemmetadata/SystemMetadataManager.java [hunk 3]
```diff
@@ -664,87 +832,81 @@ public static int convertPermission(Permission permission) {
...
-                            + " we can't move the id from the control list (lockedIds) since "
-                            + e.getMessage());
-                }
+            if(checkLock && !lockedIds.contains(guid.getValue())) {
+                throw new ServiceFailure(
+                    "0000",
+                    "The pid " + guid.getValue() + " is not locked before " + "deleting its "
+                        + "systemmetadata. There is a bug in the code. The "
+                        + "SystemMetadataManager"
+                        + ".lock" + " method must be called before call the delete method");
```
artntek, line 856: wrapping resulted in 2 unnecessary string concats

src/edu/ucsb/nceas/metacat/admin/upgrade/HashStoreUpgrader.java
artntek, line 1: Random thought: After conversion is complete, should we change the directory names to show they have been processed? This might help avoid confusion ("do I need to keep these? can I delete them??").

artntek (changes requested):
My goodness, sir - this was a very big PR - you have been busy 🤯 😁 

A couple of additional comments:

* I tested this with my local metacat instance, and it was surprisingly slow: it took about 15 - 20 mins to convert the following:
    - 87M ./data
    - 125M ./documents
    - 3.6M ./inline-data

That doesn't augur well for the ADC conversion (63T ./data; 
3.9G ./documents 😓)

* We should also test to make sure that your k8s initializer code works correctly in that environment. I will try to do that this week before I leave

src/edu/ucsb/nceas/metacat/admin/MetacatAdmin.java
```diff
@@ -46,54 +17,64 @@ public abstract class MetacatAdmin {
...
-	                }
-	            }
-	        }
-	        if (success) {
-	            PropertyService.setPropertyNoPersist("configutil.upgrade.status", SUCCESS);
-	        }
-	    } else if (status.equals(FAILURE) || status.equals(IN_PROGRESS)) {
-            //this sub upgrade process failed or is in progress, so the whole process will have the same status as well
+    public final static String IN_PROGRESS = "in progress";
+    public final static String FAILED = "failed";
```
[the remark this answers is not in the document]
taojing2002: This is a good suggestion. I hesitate to remove `FAILURE` since it has been for long time.  `FAILED` is introduced in this version. Maybe we remove it?

src/edu/ucsb/nceas/metacat/admin/upgrade/HashStoreUpgrader.java
```diff
@@ -185,25 +186,22 @@ public boolean upgrade() throws AdminException {
...
                                         });
                                         futures.add(future);
-                                        index++;
-                                        if (index == maxListLength) {
-                                            //When it reaches the max length, we need to check
-                                            // the status in the list, then reset the list
-                                            checkFuturesStatus(futures);
-                                            futures = new ArrayList<>();
-                                            logMetacat.debug("Metacat reset the future list after"
-                                                                 + " checking futures' status");
```
artntek: A suggestion to insert here. The logic is like this:

1. we start with no limit on the Set of futures
2. each time we add a future, we check to see how much memory is available
3. when the available memory drops below a certain level (say 20% of all available memory), then we use the current Set size as the upper limit
4. After this we stop checking the memory, and just use the upper limit, like you do now.

This is how it looks:

```java
// line 66 - change to:
    private int maxListLength = -1;


// ...then from line 188 onwards:
    futures.add(future);

    if (maxListLength < 0) {
        maxListLength = calculateMaxListLength(futures);
    }

    if (maxListLength > 0 && futures.size() >= maxListLength) {
        //When it reaches the max length, we need to remove
        // the complete futures from the set. So the set can
        // be reused again.
        removeCompleteFuture(futures);
    }
```

...where

```java
    private int calculateMaxListLength(Set<Future> futures) {

        // As soon as available memory becomes less than 20% of total
        // memory, use the current Set size as the upper limit
        if (Runtime.getRuntime().freeMemory() < 0.2 * Runtime.getRuntime().totalMemory()) {
            maxListLength = futures.size();
            return maxListLength;

        } else if (futures.size() > SAFETY_LIMIT) {
            //just in case :-) 
            return SAFETY_LIMIT;
        }
        return -1;
    }
```

obviously, it doesn't have to be `0.2` (20%) - could be anything you want

What do you think?

artntek (changes requested):

lib/metacat.properties
```diff
@@ -93,6 +93,8 @@ storage.hashstore.defaultNamespace=https://ns.dataone.org/service/types/v2.0#Sys
 storage.hashstore.fileNameAlgorithm=SHA-256
 storage.hashstore.directory.width=2
 storage.hashstore.directory.depth=3
+# The time out for waiting the completness of the conversion jobs. The unit is day.
+storage.hashstore.converterTimeout=5
```
artntek, line 97: not critical - but it's usually nice to include the units in the name, so there's no chance of confusion (e.g. `storage.hashstore.converterTimeoutDays`)

src/edu/ucsb/nceas/metacat/startup/MetacatInitializer.java
```diff
@@ -89,14 +89,13 @@ public void contextInitialized(ServletContextEvent sce) {
             // Check to see if the user has requested to bypass configuration
             // (dev option) and check see if metacat has been configured.
             // If both are false then stop the initialization
-            if (!ConfigurationUtil.bypassConfiguration() &&
-                                    !ConfigurationUtil.isMetacatConfigured()) {
+            if (!Boolean.parseBoolean(System.getenv("METACAT_IN_K8S"))
```
artntek, line 92: this "should" work, but someone needs to test it. I can do that when I get back, if you want? If so, could you please create an Issue, assign me, and send me the link?

src/edu/ucsb/nceas/metacat/admin/upgrade/HashStoreUpgrader.java
```diff
@@ -40,39 +40,36 @@
...
+import java.util.concurrent.TimeUnit;
 
 /**
  * @author Tao
  * This class will upgrade the old metacat file storage (/var/metacat/documents and
  * /var/metacat/data) into the HashStore storage.
  */
 public class HashStoreUpgrader implements UpgradeUtilityInterface {
-    private final static int MAX_ARRAY_LENGTH = 10000;
+    private final static int TIME_OUT = 5;
```
artntek, line 54: `TIME_OUT_DAYS`?

src/edu/ucsb/nceas/metacat/admin/upgrade/HashStoreUpgrader.java
```diff
@@ -180,18 +176,11 @@ public boolean upgrade() throws AdminException {
                                                              + " and file path(may be null): "
                                                              + path.toString());
                                         Future<?> future = executor.submit(() -> {
```
artntek, line 178: `Future` is not used any more - could remove this and delete the `import`

src/edu/ucsb/nceas/metacat/admin/upgrade/HashStoreUpgrader.java
```diff
@@ -391,14 +342,16 @@ private void handleErrorFile(File file, StringBuffer buffer) {
...
-    private void convert(Path path, String finalId, InputStream sysMetaInput, String checksum,
+    private void convert(Path path, String finalId, SystemMetadata sysMeta, String checksum,
                          String algorithm, BufferedWriter nonMatchingChecksumWriter,
                          BufferedWriter noSuchAlgorithmWriter,
                          BufferedWriter generalWriter, BufferedWriter savingChecksumTableWriter) {
         ObjectMetadata metadata = null;
         try {
-            metadata =
-                converter.convert(path, finalId, sysMetaInput, checksum, algorithm);
+            try (InputStream sysMetaInput = convertSystemMetadata(sysMeta)) {
```
artntek, line 351: I don't think you need another nested `try` here - you can just add the `(InputStream sysMetaInput = convertSystemMetadata(sysMeta))` to the first `try` statement

artntek (approved):
Looks good! I flagged a couple of things, but approved it anyway, so I'm not blocking you

PR #1979, opened 2024-10-03
Convert the objects without system metadata

taojing2002:
Now the hashstore converter can handle the objects which don't have records in the `systemmetadata` and `identifier` tables. In theory, any objects in Metacat 3.0.0 should have the systemmetadata. However, the process to generate system metadata may not completely done during the previous upgrade. So the legacy Metacat instance may have some objects without system metadata. The feature can still convert them to hashstore and generate system metadata for them.


doulikecookiedough (approved):

doulikecookiedough:
Connected with Jing on Slack to discuss misc. questions

PR #2006, opened 2024-10-30
Use heredoc to execute the multiple sql commands 

taojing2002:
Use heredoc to execute the multiple sql commands to make sure that it is a transaction.

### DIFF
--- a/src/scripts/sql/backup-restore-checksums-table/backup-checksums-table.sh
+++ b/src/scripts/sql/backup-restore-checksums-table/backup-checksums-table.sh
@@ -1,11 +1,17 @@
 #!/bin/bash
-if [ $# -ne 2 ]; then
-    echo "Usage: $0 <databaseName> <backupDirectory>"
-    exit 1
-fi
-databaseName=$1
-targetDir=$2
-fileName="checksums-backup.txt"
-path="$targetDir/$fileName"
-psql  -d $databaseName -c "\COPY (SELECT * FROM checksums) TO $path"
+
+## Script to back up the contents of the checksums table. Run this in the metacat pod. It should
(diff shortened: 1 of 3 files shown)


artntek:
I added a commit 9d89099528094a72a9feccd75e184ba68d29cd9f to allow us to run these scripts in the metacat pod, so it automatically pulls the settings it needs from the environment and the configMap

artntek (approved):

artntek:
@taojing2002 - I approved, based on your parts. 

I am not able to add you as a reviewer for my edits (because it's your PR :-) -- so please take a look at my code and make sure you don't foresee any problems. I've tested both scripts in k8s successfully.

Thanks!

taojing2002:
It looks good to me. Did you double check if the script restores the things
it drops on the table?

Thanks,

Jing

artntek:
> It looks good to me. Did you double check if the script restores the things it drops on the table? Thanks, Jing

Yep - all there: 

```sql
metacat=> \d checksums

                                          Table "public.checksums"
       Column       |          Type          | Collation | Nullable |                Default
--------------------+------------------------+-----------+----------+---------------------------------------
 checksum_id        | bigint                 |           |          | nextval('checksums_id_seq'::regclass)
 guid               | text                   |           | not null |
 checksum           | character varying(512) |           | not null |
 checksum_algorithm | character varying(250) |           | not null |

Indexes:
    "checksums_checksum" btree (checksum)
    "checksums_checksum_algorithm" btree (checksum_algorithm)
    "checksums_guid" btree (guid)

Foreign-key constraints:
    "checksums_fk" FOREIGN KEY (guid) REFERENCES systemmetadata(guid) DEFERRABLE
```

Issue #1990, opened 2024-10-22
Write a script to get delta of datasets and index them


artntek:
See #1984 

For moving deployments from legacy to k8s, we rsync the data to cephfs, then rsync again just before release, in order to get the datasets that were modified after the previous rsync.

To minimize downtime, we want to index this "delta" of datasets (instead of re-indexing all) - so we need need a script that:
1. gets the `pid` of each dataset modified since a given date/time, and then
2. calls the metacat admin API to reindex those specific datasets only, by `pid`.

mbjones:
Are you rsyncing the original `data` and `documents` directories from `/var/metacat`, or the converted hashstore directories? If the former, the filename is a mangled version of the PID for each object. So, simply saving the list of files transferred by rsync and then unmangling the PIDs woud be the list of objects that need to be indexed. If the latter, then we'd probably need to look up the PIDs for each transferred object from hashstore's PID metadata. Either way, once you have the list of PIDs, is that sufficient?

artntek:
Additional potentially-useful info, from [k8s-cluster-config/MetacatQuickRef.md](https://github.nceas.ucsb.edu/NCEAS/k8s-cluster-config/blob/04aa852d0c27b22ad233ccddb21363dd737ecdfd/MetacatQuickRef.md?plain=1#L140):

API call to list objects in reverse order of modification:
```
https://test.arcticdata.io/metacat/d1/mn/v2/object
```
API call to see if an object exists on the k8s instance yet:
```
https://arctic-dev.test.dataone.org/metacat/d1/mn/v2/object/<pid>
```
To index a list of objects:
```shell
$   TOKEN=$( kubectl get secret MYRELEASE-indexer-token \
        -o jsonpath="{.data.DataONEauthToken}" | base64 -d )

## Assuming pidsToReindex.txt contains a list of the identifiers to be indexed...
$   for pid in $(cat pidsToReindex.txt); do \
        curl -X PUT -H "Authorization: Bearer $TOKEN" \
            "https://MYHOST/MYCONTEXT/d1/mn/v2/index?pid=$pid"; \
    done
```

taojing2002:
Matthew:

My understanding is that the script only figures out the new added objects.
However, an object needs to be indexed as well even if it is not a newly
added object but its system metadata was changed. So maybe directly looking
up the modification date on the system metadata table is the fastest way.

Jing

PR #2025, opened 2024-11-21
script to get modified objects and index them

artntek:
Script for use when moving legacy installations to k8s. Gets objects modified since a certain date, and indexes them

see #1990

### DIFF
--- a/src/scripts/bash/k8s/index-delta.sh
+++ b/src/scripts/bash/k8s/index-delta.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+# Get the delta of objects created or updated since a certain time, and index them
+# See Issue #1990: https://github.com/NCEAS/metacat/issues/1990
+#
+# Use case: moving a Metacat instance to a Kubernetes cluster, we rsync the data initially to
+# cephfs (at time 't'), then do all the setup and indexing while the legacy metacat is still
+# online. Then, some time later (t + dt), we make the legacy host read-only, and do a "top-up" rsync
+# again to cephfs. We then need to index only the objects that have been created or updated between
+# t and t + dt.
+#
+
+cleanup() {
+    rm -f "${cert_file}"
+    rm -f "${result_file}"
+
+}
+
+# Initialize vars
+
+if [ $# -ne 2 ]; then
+    echo "Usage: $0 <start-time>"
+    echo "       where <start-time> is the time an hour or more before the previous rsync,"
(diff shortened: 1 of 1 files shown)


src/scripts/bash/k8s/index-delta.sh
```diff
@@ -0,0 +1,96 @@
...
+    echo "Response:"
+    curl_result=$(curl  -s  -X  PUT  --cert ${cert_file}  "${api_url}?pid=${pid}")
+    echo "${curl_result}"
+    echo
+    if [ "$(echo "${curl_result}" | grep -cz "<scheduled>\s*true\s*</scheduled>")" -lt 1 ]; then
+        echo
+        echo "Error: indexing failed -- exiting..."
+        echo
+        cleanup
+        exit 2
```
taojing2002, line 92: Is it possible to write the failed pid into a file and the index still keeps doing the job? Later we can investigate the failed pids.  So one failure will not halt the whole process.

src/scripts/bash/k8s/index-delta.sh
```diff
@@ -0,0 +1,96 @@
...
+# Now, index the objects
+result_count=$(cat ${result_file} | wc -w | xargs)
+echo "Found ${result_count} objects created/modified since ${start_time}, in release"
+echo "${release_name}."
+echo
+echo "Calling ${api_url} to index each. Enter to continue or Ctrl+C to abort..."
+read
+
+idx=0
+for pid in $(cat ${result_file}); do
```
taojing2002, line 79: Is it possible to put two or thee pids into one curl command so we can cut some network traffics? It depends on how many pids need to be indexed. If it is just dozens, maybe it is not worth to do it.

mbjones: If there are a lot of PIDs, `parallel` or `xargs -j` would likely be much faster than the serial loop. See: https://learning.nceas.ucsb.edu/2024-03-arctic/sections/parallel-programming.html#parallel-processing-in-the-shell

artntek: Thanks. The requests are very fast (they just queue the pids for indexing then return), and I don't foresee there being enough pids to warrant parallelizing (eg for ADC, there were just over 3k pids to reindex in the last month).

artntek: batched in groups of 5 pids per call

src/scripts/bash/k8s/index-delta.sh
artntek: made this change

taojing2002 (approved):

PR #2070, opened 2025-02-12
Adding tomcat ceph mount systemd config to install doc (#2069)

### LINKED ISSUE #2069: Running with Ceph mount
We have learned that when Metacat is located on a Ceph mount, Tomcat will fail to start when the system is rebooted because the mount is not ready at runtime. This can be solved by adding the Ceph mount location to the `RequiresMountsFor=` line in the `[Unit]` section of the Systemd service file.

The metacat documentation and upgrade notes should add that the following line should be in the `/etc/systemd/system/tomcat9.service` file under `[Unit]`:

```
RequiresMountsFor=/var/log/tomcat9 /var/lib/tomcat9 /var/repos/{REPO}
```

iannesbitt:
This PR closes #2069

### DIFF
--- a/docs/user/metacat/source/contributors.rst
+++ b/docs/user/metacat/source/contributors.rst
@@ -11,8 +11,8 @@ Current Contributors
   - Jing Tao (tao@nceas.ucsb.edu) [ORCID: 0000-0002-1209-5268](https://orcid.org/0000-0002-1209-5268)
   - Matthew Brooke (brooke@nceas.ucsb.edu) [ORCID: 0000-0002-1472-913X](https://orcid.org/0000-0002-1472-913X)
   - Dou Mok (mok@nceas.ucsb.edu) [ORCID: 0000-0002-6076-8092](https://orcid.org/0000-0002-6076-8092)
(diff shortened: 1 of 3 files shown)


docs/user/metacat/source/contributors.rst
```diff
@@ -36,6 +36,7 @@ Previous Contributors
   - Peter Slaughter (slaughter@nceas.ucsb.edu)
   - Rob Nahf (rnahf@epscor.unm.edu)
   - Matthew Brooke (brooke@nceas.ucsb.edu)
+  - Ian Nesbitt (nesbitt@nceas.ucsb.edu)
```
artntek, line 39: should this be moved up to "Current Contributors"?

artntek (approved):
approved, with a small change suggestion

docs/user/metacat/source/contributors.rst
mbjones: yes, I think it should be in current contributors (and added to our citation info and contributors list in the README too)

iannesbitt: Ok, done. I put myself in before Matt (who I assume is last as the PI).

Issue #2226, opened 2025-08-20
Use CloudNative PG Operator Instead of Bitnami PostgreSQL Chart


artntek:
see #2218 and https://github.com/DataONEorg/k8s-cluster/issues/65

artntek:
There is also a [dataone-indexer dependency](https://github.com/DataONEorg/dataone-indexer/issues/281)

artntek:
### Notes for disabling SSL on client side, for postgres connections

https://jdbc.postgresql.org/documentation/use/#connection-parameters
```java
props.setProperty("ssl", "false");
```
or:
https://jdbc.postgresql.org/documentation/ssl/#using-ssl-without-certificate-validation
```java
props.setProperty("sslfactory", "org.postgresql.ssl.NonValidatingFactory");
```
I think setting the props is equivalent to adding the params on the end or your jdbc connection string like `.../metacat?sslmode=disable`  (see [configuring the client](https://jdbc.postgresql.org/documentation/ssl/#configuring-the-client) for difference between `ssl=false` and `sslmode=disable`)

artntek:
## Data Import at Startup: Notes on Progress/Findings

In broad terms, CNPG can `bootstrap` a new cluster using data from elsewhere, in these ways:
1. **Importing data** via an automated `pg_dump`/`pg_restore` process 
   - @jeanetteclark used this successfully for metadig
   - Works even when source postgres major version < CNPG postgres version
2. **Streaming replication** from another PG instance, using `pg_basebackup`
   - Got this to work for metacat (see below)
   - Works only across equal major versions of PG
   - However, upon completion, immediately promotes CNPG and stops replication
3. **Using Recovery data** from volume snapshots or from an Object store (Barman Cloud Plugin)
   - Volume snapshot recovery seems like a useful feature for backups, since Ceph-CSI can do snapshots

> [!NOTE]
> The problem with all of these is that they require Metacat to be in Read-Only mode for the entire bootstrap (otherwise any Bitnami writes made during cut-over would not be synced).
> 
> Replication seems to be the most promising approach for least downtime/RO time, but if used for bootstrapping, it is automatically stopped by CNPG upon completion - so new writes to Bitnami PG after that would be lost (unless the operator continuously monitors, and puts Metacat in RO mode right before replication is completed).
>
> Next step, therefore, is to look at using continuous replication from Bitnami PG, instead of a one-and-done bootstrap.


### Replication Gotchas:

[...]

PR #2229, opened 2025-09-03
Use cnpg instead of bitnami postgresql

artntek:
see #2226


mbjones:
@vchendrix This refactor of our Metacat helm chart to use CloudNativePG may be of interest to you...

helm/templates/statefulset.yaml
```diff
@@ -30,7 +30,7 @@ spec:
               ## wait for PostgreSQL to be ready...
               ##
               PGHOST={{ include "metacat.postgresql.hostname" . }};
-              PGPORT={{ .Values.postgresql.primary.service.ports.postgresql -}};
+              PGPORT=5432;
```
mbjones, line 33: Isn't this configurable, and so why did it change to hardcoded in the stateful set? Wouldn't it be good to set it, especially if a non-CNPG instance is used?

helm/values.yaml
```diff
@@ -1020,151 +992,68 @@ readinessProbe:
...
-
-    ## Postgresql persistence
-    ## For explanatory notes, see top-level `persistence` section for metacat, and the bitnami
-    ## postgresql README at https://github.com/bitnami/charts/blob/main/bitnami/postgresql/README.md
-    ## @param postgresql.primary.persistence.enabled Enable data persistence using PVC
-    ## @param postgresql.primary.persistence.existingClaim Existing PVC to re-use
-    ## @param postgresql.primary.persistence.storageClass Storage class of backing PV
-    ## @param postgresql.primary.persistence.size PVC Storage Request for postgres volume
+    pg_ident:
+      - local       postgres       metacat
```
mbjones, line 1038: Does this map the postgres user to the metacat user, or vice versa? I'm not sure this mapping is a good idea because postgres is an admin account, but let's discuss how it works and what exposure this might cause..

mbjones (commented):
Looks great, excited to see the nice postgres CNPG template. I think it would be useful to discuss the ident mappings and the pg_hba.conf before you release. Nicely done!

helm/templates/statefulset.yaml
artntek: you're right. I got carried away reading the CNPG docs (always 5432 and not configurable there, afaict), but yes - should be configurable to allow for alternative PG setup.

This also begs a bigger question; I'd put postgres-related stuff under `.Values.cnpg...`, which I guess is fine for CNPG-specific settings, but I guess it would be better to use a more generic `.Values.database...` hierarchy, in case CNPG's not used

artntek: reverted in d572ce17b073e32544325e2a22720adae5870314

helm/values.yaml
mbjones: ok, that's fine. the `all all trust` model doesn't follow the principle of least privilege, so we should discuss that sometime, but if your new rule just does what we used to do, then that is fine for now. Given that the password is in the secret anyways, anyone that can exec into the container can get the pw anyway so it probably makes little difference.

PR #2233, opened 2025-09-05
Bumped commons-io to 2.20.0.

taojing2002:
It also added code to download multiple versions of junit-bom in pom.xml.


doulikecookiedough (approved):
Looks good!

Issue #2277, opened 2026-01-27
Make Original Client IP Available to Metacat in K8s


artntek:
In K8s, the ingress now provides the real (remote) client IP address in the `x-real-ip` and `x-forwarded-for` headers. However, tomcat still logs the (local) IP address of the ingress itself (192.x.x.x).

Tomcat can be configured to use the real IP, instead, though; see:
- https://stackoverflow.com/a/55275335
- https://tomcat.apache.org/tomcat-9.0-doc/config/valve.html#Access_Log_Valve
- https://tomcat.apache.org/tomcat-9.0-doc/config/valve.html#Remote_Ip_Valve

This requires editing server.xml and including it in the docker image

artntek:
released in metacat 3.4.2/chart-4.2.2

PR #2278, opened 2026-01-27
use x-forwarded-for header as client IP

artntek:
see #2277


mbjones:
Is there a blocker on this PR @artntek ? Seems like it has been done for a while, and would be useful in production.

artntek:
no blocker, just a constant onslaught of new issues that are always more important than everything else :-) I agree it would be great to merge this- I just need to pick up the context again and do some testing

taojing2002 (approved):

Issue #2158, opened 2025-05-08
Create Python submitter for CN Dataone-indexer 


taojing2002:
We will use the postgres sql command to create a trigger on the systemmetadata table. So the cn-index-generator can listen the notification from the trigger. This is part the task to [Integrate Dataone-indexer to Dataone CN](https://github.com/DataONEorg/dataone-indexer/issues/189)

taojing2002:
A [link](https://gist.github.com/goliatone/5fbeb1912e5937e8e3cf94618be9bebf) to describe how to set up the table events notification in PostgreSQL.

taojing2002:
The code and documentation were reviewed and merged into the develop branch.

PR #2281, opened 2026-02-19
Python index-tasks submitter for CN

taojing2002:
The branch accommodates the usage of DataONE-Indexers for CNs.
It has two clients:
1. Pulling new records from the system metadata - the regular submitter.
2. Delete events listener - the submitter for the delete events

It also has a utility script - reindex.py to reindex the missing objects.


artntek:
see Issue #2158

artntek (approved):

Issue #1709, opened 2023-10-02
Support HTTP Range Requests in MNRead.get


robyngit:
Detect and handle HTTP [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) to enable clients to retrieve a portion of a file without the need to download the entire content. This feature would allow MetacatUI and other clients to [preview data files before downloading them](https://github.com/NCEAS/metacatui/issues/1758). It would also allow clients to resume downloads in the event of a network interruption.

- @mbjones mentioned the possibility of implementing this without making changes to the DataONE API: A range request could be made via HTTP headers, leaving the request body unchanged and having Metacat only handle the range request headers.
- It would be the client's responsibility to generate the range request, for example:
```bash
curl -H "Range: bytes=0-1000" https://dev.nceas.ucsb.edu/knb/d1/mn/v2/object/urn%3Auuid%3A24b85258-3e86-40cb-accc-28153513dea8
```
- The feature could be a non-mandatory enhancement, such that the existing behavior remains consistent for repositories not making use of range requests.
- Apache Tomcat and the Servlet API might provide built-in support for HTTP range requests.
- A discussion is needed on how this feature interacts with event metrics:
  - Is a range request categorized as a download or a view/read?
  - Does this require a new event type, e.g. "partial read", "preview"?

Note: `https://dev.nceas.ucsb.edu/knb/d1/mn/v2/object/urn%3Auuid%3A24b85258-3e86-40cb-accc-28153513dea8` gives a 100,000 line CSV file that could be useful for testing

taojing2002:
@robyngit   Two questions:
1. Does this feature only support the text data files (e.g, cvs)? How about Excel files?
2. What are the units of the range? Lines or bytes or both?

mbjones:
[...]

PR #2282, opened 2026-02-23
Feature 1709 partial read object

taojing2002:
Add the new feature to support retrieving partial objects. The ticket is [here](https://github.com/NCEAS/metacat/issues/1709).

### DIFF
--- a/docs/user/metacat/source/retrieving-partial-objects.rst
+++ b/docs/user/metacat/source/retrieving-partial-objects.rst
@@ -1,0 +1,30 @@
+Retrieving Partial Objects
+==========================
+
+Users can retrieve a partial object using the HTTP `Range` header with the DataONE `GET /object/{id}` API.
+
+ - **Syntax:**
+
+  ``-H "Range: bytes={start}-{end}" GET /object/{id}``
+
+ - **Rules:**
+
+  - The `start` value **must be specified**.
+  - The first byte of an object is byte `0`.
+  - `start` must be **less than or equal to** `end`.
+  - `end` must be **less than the total object size**.
+  - If `end` is omitted, the response will include all bytes from `start` to the end of the object.
+
+::
+
+   # Examples
+
+   // Get the first byte of object `foo.1`
+   curl -H "Range: bytes=0-0" -X GET https://knb.ecoinformatics.org/knb/d1/mn/v2/object/foo.1
+
+   // Get 3 bytes of object `foo.1`, starting from the second byte
+   curl -H "Range: bytes=1-3" -X GET https://knb.ecoinformatics.org/knb/d1/mn/v2/object/foo.1
+
+   // Get all bytes of object `foo.1` except the first byte
+   curl -H "Range: bytes=1-" -X GET https://knb.ecoinformatics.org/knb/d1/mn/v2/object/foo.1
+
--- a/docs/user/metacat/source/retrieving-partial-objects.rst
+++ b/docs/user/metacat/source/retrieving-partial-objects.rst
@@ -9,11 +9,10 @@ Users can retrieve a partial object using the HTTP `Range` header with the DataO
 
  - **Rules:**
 
-  - The `start` value **must be specified**.
-  - The first byte of an object is byte `0`.
-  - `start` must be **less than or equal to** `end`.
-  - `end` must be **less than the total object size**.
-  - If `end` is omitted, the response will include all bytes from `start` to the end of the object.
+  - The ``start`` value **must be specified**, and must be **greater than or equal** to ``0`` (since ``0`` denotes the first byte).
+  - ``start`` must be **less than or equal to** ``end``.
+  - ``end`` must be **less than the total object size**.
+  - If ``end`` is omitted, the response will include all bytes from ``start`` to the end of the object.
 
 ::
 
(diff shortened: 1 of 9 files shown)


docs/user/metacat/source/retrieving-partial-objects.rst
```diff
@@ -0,0 +1,30 @@
...
+Users can retrieve a partial object using the HTTP `Range` header with the DataONE `GET /object/{id}` API.
+
+ - **Syntax:**
+
+  ``-H "Range: bytes={start}-{end}" GET /object/{id}``
+
+ - **Rules:**
+
+  - The `start` value **must be specified**.
+  - The first byte of an object is byte `0`.
```
artntek: This might be a little more clear if we combined them, like this:

> - The `start` value **must be specified**, and must be **greater than or equal to** `0` (since `0` denotes the first byte).

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
```diff
@@ -626,4 +627,39 @@ protected void getSessionFromHeader() {
...
+
+    /**
+     * Write the bytes (either whole object file or a specified range) back to the client
+     * @param data the input stream contains the bytes of the object
+     * @throws IOException
+     * @throws InvalidRequest
+     */
+    protected void writeToResponse(InputStream data) throws IOException, InvalidRequest {
+        ByteRange range = ByteRange.parseRange(request);
+        response.setHeader("Accept-Ranges", "bytes");
```
artntek, line 640: did you intend to add this header to all responses (i.e. even those that were not requested using a byte range)?

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
```diff
@@ -626,4 +627,39 @@ protected void getSessionFromHeader() {
...
+                    IOUtils.copyLarge(data, out, 0, length);
+                } else {
+                    // Open-ended range
+                    response.setHeader("Content-Range", "bytes " + range.getStart() + "-/*");
+                    IOUtils.skipFully(data, range.getStart());
+                    IOUtils.copyLarge(data, out);
+                }
+            }
+            out.flush();
+        }
```
artntek, line 673: should we have a `finally` block that calls `IOUtils.closeQuietly()`?

(I don't know the answer - just curious, since the code you replaced had `closeQuietly` statements)

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
```diff
@@ -626,4 +627,39 @@ protected void getSessionFromHeader() {
...
+                response.setStatus(200);
+                IOUtils.copyLarge(data, out);
+            } else {
+                // Partial content
+                response.setStatus(HttpServletResponse.SC_PARTIAL_CONTENT);
+                if (range.getEnd() != null) {
+                    long length = range.getEnd().longValue() - range.getStart() + 1;
+                    response.setHeader("Content-Range",
+                                       "bytes " + range.getStart() + "-" + range.getEnd()
+                                           .longValue() + "/*");
```
artntek, line 663: should we also set the `Content-Length` header, since we have that value?

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
```diff
@@ -626,4 +627,39 @@ protected void getSessionFromHeader() {
...
+        ByteRange range = ByteRange.parseRange(request);
+        response.setHeader("Accept-Ranges", "bytes");
+        try (OutputStream out = response.getOutputStream()) {
+            if (range == null) {
+                // Full stream, size unknown
+                response.setStatus(200);
+                IOUtils.copyLarge(data, out);
+            } else {
+                // Partial content
+                response.setStatus(HttpServletResponse.SC_PARTIAL_CONTENT);
```
artntek, line 658: What happens if the start value is bigger than the length of the object? (eg if the object is 1024 bytes long, but the range request is 2000-3000?) I think in that case, we should be returning a `416 Range Not Satisfiable` (see https://www.rfc-editor.org/rfc/rfc9110.html#name-416-range-not-satisfiable)

artntek (changes requested):

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
taojing2002: We decided to keep it.

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
taojing2002: We use the try-with-resource, the output stream will be closed automatically.

src/edu/ucsb/nceas/metacat/restservice/D1ResourceHandler.java
```diff
@@ -644,18 +645,26 @@ protected void writeToResponse(InputStream data) throws IOException, InvalidRequ
                 IOUtils.copyLarge(data, out);
             } else {
                 // Partial content
+                try {
+                    IOUtils.skipFully(data, range.getStart());
+                } catch (IllegalArgumentException | EOFException e) {
+                    // Incorrect the start number
+                    response.setStatus(HttpServletResponse.SC_REQUESTED_RANGE_NOT_SATISFIABLE);
+                    throw new InvalidRequest("1010",
+                                             "Incorrect the start byte " + range.getStart() + ". "
```
artntek: correct wording should be `Incorrect start byte: ` (without ~~the~~) 

However, it would be more useful to the user to say something like:
```java
"Start byte (" + range.getStart() + ") cannot be greater than object length ("
    + length + ")"
```

(...but I'm not sure if this `catch` clause will be called for any other errors, that are different from this one?)

artntek (changes requested):

artntek (approved):

Issue #2290, opened 2026-05-01
Don't sort objects by types in the reindex-all command


taojing2002:
Previously Dataone-Indexer could only successfully index a resource map object when the metadata and data objects aggregated by the resource map have been already indexed, which means their Solr documents are available. So Metacat's reindex-all command sorts objects by types: reindex all metadata/data objects of the Metacat before start to reindex resource map objects, in order to improve the successful possibilities of indexing resource map objects. 
After the introduce of the dummy solr doc, indexing resource map doesn't have the requirement any more. However, we haven't changed the behavior of the reindex-all command - still metadata/data objects first, then resource map.
Concentration of indexing resource map objects has a draw back - more easily to cause the version conflict issue. For example, both resource map r1 and r2 aggregate the same data object d. When we index both r1 and r2 concurrently, both r1 and r2 will change the solr doc of d.  If one of the processes doesn't get the latest version solr document of d at the beginning, the version conflict happens and the resource map index process has to restart again.  If we don't sort the objects by type, the index order can be possibly changed to d, r1, other objects, r2. So r2 will be indexed much later than r1. The possibility of version conflict will decrease and the performance will be improved.

taojing2002:
The branch was reviewed and merged into develop.

PR #2292, opened 2026-05-01
Don't sort objects by resourcemap or non-resourcemap in reindex-all calls

taojing2002:
Ticket is [here.](https://github.com/NCEAS/metacat/issues/2290)

### DIFF
--- a/src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
+++ b/src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
@@ -3680,9 +3680,9 @@ public class MNodeService extends D1NodeService
      * Index all objects. We don't put the list of pids in a vector anymore.
      */
     private void buildAllObjectIndex() {
-        StringBuilder sql = new StringBuilder("select guid from systemmetadata");
+        String sql = "select guid from systemmetadata";
         try {
-            long size = buildIndexFromQuery(sql.toString());
+            long size = buildIndexFromQuery(sql);
             logMetacat.info("The number of objects is " + size
                                 + " being submitted to the index queue.");
         } catch (SQLException | ServiceFailure e) {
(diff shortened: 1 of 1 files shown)


src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
artntek, line 3667: it might be a good addition to check if the reindex is already running, and log a warning & return if so?

src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
```diff
@@ -3669,89 +3669,25 @@ protected void handleReindexAllAction() {
...
 
     /**
-     * Index all non-resourcemap objects first. We don't put the list of pids in a vector anymore.
-     * @param resourceMapFormatList  the list of the resource map format
+     * Index all objects. We don't put the list of pids in a vector anymore.
      */
-    private void buildAllNonResourceMapIndex(List<String> resourceMapFormatList) {
-        boolean firstTime = true;
+    private void buildAllObjectIndex() {
         StringBuilder sql = new StringBuilder("select guid from systemmetadata");
```
artntek: this can just be a `String`, instead of using a `StringBuilder` and then needing to call `toString()`

artntek (changes requested):
2 small suggestions

src/edu/ucsb/nceas/metacat/dataone/MNodeService.java
```diff
@@ -3665,11 +3666,21 @@ protected void handleReindexAction(List<Identifier> pids) {
      */
     protected void handleReindexAllAction() {
         // Process all of the documents
-        logMetacat.debug("MNodeService.handleReindexAllAction - "
-                                               + "reindex all objects in this Metacat instance");
+        //Check if another thread has been running the index-all command
+        //The first boolean variable is the expected value, and the second is the update value.
+        if (!indexAllProcessRunning.compareAndSet(false, true)) {
```
artntek, line 3671: This `!` combined with `compareAndSet()` is a bit difficult to understand for my poor old brain. It might be easier to read (and more safe), if we change this line to use a simple `get()` call instead, and then add a `set(true)` before line 3680. What do you think?

artntek (changes requested):

artntek (approved):

PR #2313, opened 2026-05-21
Task 2312 release 3.4.1 prep

taojing2002:


lib/web.xml
```diff
@@ -157,10 +157,10 @@
     </servlet-mapping>
 
     <!-- CN REST services v1-->
-    <servlet-mapping>
+    <!--<servlet-mapping>
```
artntek, line 162: let's add a comment here too, to help our future selves recall why it's commented out

lib/web.xml
```diff
@@ -169,10 +169,10 @@
     </servlet-mapping>
 
     <!-- CN REST services v2-->
-    <servlet-mapping>
+    <!--<servlet-mapping>
```
artntek, line 176: same as above

artntek (changes requested):

artntek (approved):

PR #2314, opened 2026-05-21
Task 2312 release 3.4.1 prep

artntek:
see #2312


mbjones (approved):
LGTM, but please include CVE numbers when known in the Security-Advisories.md bulleted metadata.

